PR #9, opened 2019-04-12
Add build scripts for app-sre deploy

jharrington22:
This PR adds the deploy scripts required to deploy via the app-sre pipeline

PR #12, opened 2019-04-12
SREP-1143: Get ready for Prow & osd-operators-registry

lisa:
Get ready for Prow & osd-operators-registry

Signed-off-by: Lisa Seelye <lseelye@redhat.com>


lisa:
This is similar to #9...

build/Dockerfile
```diff
@@ -1,17 +1,16 @@
-FROM alpine:3.8
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+ENV OPERATOR_PATH=/go/src/github.com/openshift/pagerduty-operator
+COPY . ${OPERATOR_PATH}
+WORKDIR ${OPERATOR_PATH}
+RUN make gobuild
 
-RUN apk --no-cache add ca-certificates
```
rogbas: if the pagerduty operator connects to pagerduty or other external URLs via https it might need the ca-certificates package for the Root CAs.

rogbas: If that's the case we should install it on the 2nd stage of the Dockerfile

lisa: We're not supposed to use alpine though.

rogbas: We could install it with yum, like `yum install -y ca-certificates && yum clean all`

PR #22, opened 2019-04-18
Listen on all namespaces

wshearn:
We want to listen across the entire cluster and not just in 1 namespace


wshearn:
> unit test?

that is out of scope of this PR. They are on my TODO list

jewzaam:
https://github.com/openshift/pagerduty-operator/issues/23

/lgtm

PR #24, opened 2019-04-18
Add clusterdeployments to the role

wshearn:
Add permissions that lets us set finalizers on cluster deployments


lisa:
Will this work? In the chat you mentioned that it was across namespaces, but this is a `Role` with a `RoleBinding`, which don't cross namespaces.

PR #26, opened 2019-04-18
Migrate to unique PagerDuty service names

wshearn:
Make the PagerDuty Service name unique across all hive instances.
On SyncSet delete try and get the service first, if that fails create a
new PagerDuty Service.

We need to make sure that the PagerDuty Service names are unique across all hive instances. Also on SyncSet delete we should try running a GetService first, if that fails run a CreateService. This  is useful in the initial migration to unique service names

/cc @jharrington22 
/cc @jewzaam

### DIFF
--- a/pkg/controller/syncset/syncset_deleted.go
+++ b/pkg/controller/syncset/syncset_deleted.go
@@ -18,20 +18,41 @@ import (
 	"context"
 	"strings"
 
+	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controller/syncset/syncset_deleted.go
```diff
@@ -18,20 +18,41 @@ import (
...
+
+	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: cdName}, clusterdeployment)
+	if err != nil {
+		// Error finding the cluster deployment, requeue
+		return reconcile.Result{}, err
+	}
+
 	pdData := &pd.Data{
-		ClusterID: strings.Split(request.Name, "-")[0],
+		ClusterID:  clusterdeployment.Name,
```
jewzaam:
```suggestion
		ClusterID:  clusterdeployment.Spec.ClusterName,
```

wshearn: Should this be applied everywhere? Because in already commited code I am just using the CD name

PR #27, opened 2019-04-18
Fix a couple edge cases

wshearn:
During migration to the new pagerduty service name I hit a few edge cases. This should solve them.

Edge Cases Fixed:

- If the syncset is deleted and a new operator is rolled out, it does not know how to handle reconciling them since they have the finalizer, I changed how I set the finalizer, and then I run a GetService instead of just create, so it will check to see if any existing PD Services exist and reuse that before creating a new service

- Fixed the case of the CD name being different from the clusterid in the syncset. It seems that the cluster deployment does not need to share the same name as the cluster id, so I am passing in the correct name for the ClusterID to the syncset.

/cc @jewzaam 
/cc @lisa

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_created.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_created.go
@@ -20,20 +20,19 @@ import (
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	hivecontrollerutils "github.com/openshift/hive/pkg/controller/utils"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/pagerduty/syncset.go
```diff
@@ -34,7 +34,7 @@ func (data *Data) GenerateSyncSet(namespace string, name string, pdServiceID str
 		Spec: hivev1.SyncSetSpec{
 			ClusterDeploymentRefs: []corev1.LocalObjectReference{
 				{
-					Name: name,
+					Name: data.ClusterID,
```
jewzaam: Think the `ClusterDeploymentRefs` value needs to be metadata.name, which is passed as `name` param to this function.  The ClusterID is being set to the `ClusterID` property in the ClusterDeployment spec.

PR #34, opened 2019-05-02
Create Integrations for each new service

wshearn:
Create Integrations for each service created and use the integration key inside the syncset

/cc @lisa 
/cc @jewzaam

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
@@ -16,11 +16,11 @@ package clusterdeployment
 
 import (
 	"context"
-	"fmt"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	hivecontrollerutils "github.com/openshift/hive/pkg/controller/utils"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -29,7 +29,7 @@ import (
 func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, instance *hivev1.ClusterDeployment) (reconcile.Result, error) {
 	r.reqLogger.Info("Deleting syncset")
 
-	ssName := fmt.Sprintf("%v-pd-sync", instance.Name)
+	ssName := instance.Name + "-pd-sync"
 	ss := &hivev1.SyncSet{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: ssName, Namespace: request.Namespace}, ss)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/pagerduty/service.go
```diff
@@ -108,22 +108,44 @@ func (data *Data) ParsePDConfig(osc client.Client) error {
...
-		return "", errors.New("Multiple services returned from PagerDuty")
+		return pdApi.Service{}, errors.New("Multiple services returned from PagerDuty")
 	}
 
-	return services.Services[0].ID, nil
+	return services.Services[0], nil
+}
+
+// GetIntegrationKey searches the PD API for an already existing service and returns the first integration key
+func (data *Data) GetIntegrationKey() (string, error) {
```
lisa: What does accessing the first integration key get us? And if there's more than one, what does that mean; does this method operate on a specific type of integration (eg DeadManSnitch) only?

pkg/pagerduty/service.go
```diff
@@ -141,25 +163,37 @@ func (data *Data) CreateService() (string, error) {
...
 	}
 
 	newSvc, err := client.CreateService(clusterService)
 	if err != nil {
 		return "", err
 	}
 
-	return newSvc.ID, nil
+	clusterIntegration := pdApi.Integration{
+		Name: "Test",
```
lisa: Surely this isn't intended? Or is this PR a WIP?

pkg/pagerduty/service.go
wshearn: DeadManSnitch will have it's own service. The problem is that integrations can have the same name, so we don't know if it is the proper one, if we throw an error I plan on adding metrics that track the error count so if we detect an issue we can go look. https://i.imgur.com/f0pcfGM.jpg

pkg/pagerduty/service.go
wshearn: I had left that there while I try and figure out a better name, no idea what to call it.

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -29,7 +29,7 @@ import (
 func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, instance *hivev1.ClusterDeployment) (reconcile.Result, error) {
 	r.reqLogger.Info("Deleting syncset")
 
-	ssName := fmt.Sprintf("%v-pd-sync", instance.Name)
+	ssName := instance.Name + "-pd-sync"
```
jharrington22: Why switch this?

wshearn: After reading it all over it was just easier for me to read

pkg/kube/configmap.go
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+package kube
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// GenerateConfigMap returns a syncset that can be created with the oc client
```
lisa: This comment doesn't match up with the function. The function is returning a `*corev1.ConfigMap` but the comment seems to suggest a SyncSet would be returned.

PR #40, opened 2019-05-17
Deploy needed objects for prometheus metrics when the operator comes up

fahlmant:
This PR ensures that when the operator is deployed, all the proper objects are placed so prometheus can scrape the custom metrics.

-The operator creates a pagerduty-operator service to expose the metrics over on deploy
-The operator creates a pagerduty-operator servicemonitor to tell prometheus how to digest the metrics
-Removes the operator-sdk metrics service for now
-Includes a prometheus-k8s role for the pagerduty-operator namespace that has permissions to get, list and watch services, pods and endpoints.
-Includes a prometheus-k8s rolebinding to allow the prometheus-k8s serviceaccount from the openshift-monitoring namespace to use the aformentioned role.

When the operator and the role/rolebindings are deployed, the pagerduty-operator namespace must be given the following label to get everything working: openshift.io/cluster-monitoring="true"

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -23,11 +23,13 @@ import (
 
 	"github.com/openshift/pagerduty-operator/pkg/apis"
 	"github.com/openshift/pagerduty-operator/pkg/controller"
+	operatormetrics "github.com/openshift/pagerduty-operator/pkg/metrics"
 
 	hivev1 "github.com/openshift/hive/pkg/apis/hive/v1alpha1"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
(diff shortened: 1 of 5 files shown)


lisa:
> -Includes a prometheus-k8s role for the pagerduty-operator namespace that has permissions to get, list and watch services, pods and endpoints.
> -Includes a prometheus-k8s rolebinding to allow the prometheus-k8s serviceaccount from the openshift-monitoring namespace to use the aformentioned role.

Why does the prometheus-k8s service account need to use that role?

fahlmant:
@lisa The prometheus serviceaccount in openshift-monitoring does not have proper permissions to scrape the service by default. @rogbas helped diagnose and write the role/rolebinding

lisa:
@fahlmant Oh, that makes sense. I was thinking in terms of that serviceaccount interacting in a different way!

/lgtm

PR #42, opened 2019-05-20
Add prometheus role and rolebinding to the generated CSV

fahlmant:
This PR adds the needed prometheus-k8s RBAC objects to the generation of the CSV


fahlmant:
@jharrington22 Could you give this a look over?

wshearn:
these should have been numbered 05 and 06 and have operator renamed to 07. They are numbered in the order they are supposed to be applied

fahlmant:
Technically these don't need to be deployed before the operator. These don't give the operator any permissions, but allows the prometheus-k8s serviceaccount from the `openshift-monitoring` namespace to look at the metrics given by the operator. I can rename them, but the operator would still come up fine if these were added after.

PR #48, opened 2019-06-20
Remove ServiceMonitor in favor of Route for external scraping

fahlmant:
This PR updates the local metrics package to the latest method. It also replaces the creation of a ServiceMonitor with a Route to enable external scraping. 

For reference, the metrics pkg code already exists in the aws-account-operator: https://github.com/openshift/aws-account-operator/blob/master/cmd/manager/main.go

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,7 +29,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 
-	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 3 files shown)


cmd/manager/main.go
```diff
@@ -125,25 +125,11 @@ func main() {
...
-	} else {
-		log.Info("Created Service")
-		err = mgr.GetClient().Create(context.TODO(), sm)
-		if err != nil {
-			log.Error(err, "error creating metrics ServiceMonitor")
-		} else {
-			log.Info("Created ServiceMonitor")
-		}
+	// Configure metrics if it errors log the error but continue
+	if err := operatormetrics.ConfigureMetrics(context.TODO()); err != nil {
```
jharrington22: Does the service monitor do this step for us? How come we did not do it beofre?

PR #49, opened 2019-06-26
SREP-1352: Specify an IncidentUrgencyRule for new PD Services

lisa:
PagerDuty API has the ability to map incident severity to urgency. By
default PagerDuty treats all incidents (even low severity ones) as high
urgency incidents and thus will page, even for the most minor warnings.
This behaviour isn't desired (See SREP-1352). This change will have new
PagerDuty services (eg new clusters) created with a `severity_based`
Incident Urgency Rule to instruct PagerDuty to defer to the incident's
severity to inform urgency.

Signed-off-by: Lisa Seelye <lseelye@redhat.com>


lisa:
This is ready for 👀 and can be merged, having been tested.

PR #50, opened 2019-07-18
Metrics

liamwazherealso:
https://jira.coreos.com/browse/SREP-1297

This PR has two commits. One with the changes that will be eventually merged. The other with temp changes for testing on a v4 cluster.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,8 +25,12 @@ import (
 	"github.com/openshift/operator-custom-metrics/pkg/metrics"
 	"github.com/openshift/pagerduty-operator/pkg/apis"
 	"github.com/openshift/pagerduty-operator/pkg/controller"
+	"github.com/openshift/pagerduty-operator/pkg/localmetrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	routev1 "github.com/openshift/api/route/v1"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
@@ -124,11 +128,29 @@ func main() {
 	}
 
 	metricsServer := metrics.NewBuilder().WithPort(metricsPort).WithPath(metricsPath).
+		WithCollectors(localmetrics.MetricPagerDutyCreateFailure).
+		WithCollectors(localmetrics.MetricPagerDutyDeleteFailure).
+		WithCollectors(localmetrics.MetricPagerDutyHeartbeat).
 		GetConfig()
+
 	// Configure metrics if it errors log the error but continue
+
 	if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
 		log.Error(err, "Failed to configure Metrics")
 	}
+
+	client := mgr.GetClient()
+	pdAPISecret := &corev1.Secret{}
+	err = client.Get(context.TODO(), types.NamespacedName{Namespace: "pagerduty-operator", Name: "pagerduty-api-key"}, pdAPISecret)
+	var APIkey string
+
+	if err == nil {
+		APIkey = string(pdAPISecret.Data["PAGERDUTY_API_KEY"])
+	} else {
+		log.Error(err, "Failed to get secret")
+	}
+
+	go localmetrics.UpdateAPIMetrics(APIkey)
 
 	log.Info("Starting the Cmd.")
 
--- a/pkg/controller/clusterdeployment/clusterdeployment_created.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_created.go
@@ -21,6 +21,7 @@ import (
 	hivecontrollerutils "github.com/openshift/hive/pkg/controller/utils"
 	"github.com/openshift/pagerduty-operator/config"
 	"github.com/openshift/pagerduty-operator/pkg/kube"
+	localmetrics "github.com/openshift/pagerduty-operator/pkg/localmetrics"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -41,6 +42,8 @@ func (r *ReconcileClusterDeployment) handleCreate(request reconcile.Request, ins
 		}
 	}
 
+	ClusterID := instance.Spec.ClusterName
+
 	pdData := &pd.Data{
 		ClusterID:  instance.Spec.ClusterName,
 		BaseDomain: instance.Spec.BaseDomain,
@@ -54,9 +57,11 @@ func (r *ReconcileClusterDeployment) handleCreate(request reconcile.Request, ins
 		var createErr error
 		pdIntegrationKey, createErr = r.pdclient.CreateService(pdData)
 		if createErr != nil {
+			localmetrics.UpdateMetricPagerDutyCreateFailure(1, ClusterID)
 			return reconcile.Result{}, createErr
 		}
 	}
+	localmetrics.UpdateMetricPagerDutyCreateFailure(0, ClusterID)
 
 	pdIntegrationKey, err = r.pdclient.GetIntegrationKey(pdData)
 	if err != nil {
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -15,22 +15,97 @@
 package localmetrics
 
 import (
+	"net/http"
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("localmetrics")
+
+const (
+	apiEndpoint = "https://api.pagerduty.com/users"
 )
 
 var (
-	metricPlaceholder = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "placeholder",
-		Help: "Placeholder",
+	MetricPagerDutyCreateFailure = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "metric_pageduty_create_failure",
+		Help: "Metric for the failure of creating a cluster deployment.",
+	}, []string{"name", "clusterdeployment_name"})
+
+	MetricPagerDutyDeleteFailure = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "metric_pageduty_delete_failure",
+		Help: "Metric for the failure of deleting a cluster deployment.",
+	}, []string{"name", "clusterdeployment_name"})
+
+	MetricPagerDutyHeartbeat = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "metric_pageduty_heartbeat",
+		Help: "Metric for heartbeating of the Pager Duty api.",
 	}, []string{"name"})
 
 	MetricsList = []prometheus.Collector{
-		metricPlaceholder,
+		MetricPagerDutyCreateFailure,
+		MetricPagerDutyDeleteFailure,
+		MetricPagerDutyHeartbeat,
 	}
 )
 
-// UpdatePlaceholderGauge ...
-func UpdatePlaceholderGauge() {
+// UpdateAPIMetrics updates all API endpoint metrics ever 5 minutes
+func UpdateAPIMetrics(APIkey string) {
+	d := time.Tick(5 * time.Minute)
+	for range d {
+		UpdateMetricPagerDutyHeartbeatGauge(APIkey)
+	}
 
-	metricPlaceholder.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(1))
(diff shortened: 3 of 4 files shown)


liamwazherealso:
Upon testing I'm getting ```{"level":"error","ts":1563480645.8196716,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"clusterdeployment-controller","request":"uhc-staging-16tlasvtfs24acir1aqfne2ukpq59j29/lseelye-srep-1659","error":"Failed call API endpoint. HTTP response code: 404. Error: &{2100 Not Found <nil>}"
```

File: manifests/05-operator.yaml [hunk 1]
```diff
@@ -16,7 +16,7 @@ spec:
       serviceAccountName: pagerduty-operator
       containers:
         - name: pagerduty-operator
-          image: quay.io/app-sre/pagerduty-operator
+          image: quay.io/lpieri/pagerduty-operator-metrics:metrics
```
fahlmant: I think you left in some testing stuff here. Please revert changes to this file

manifests/apply.sh
```diff
@@ -0,0 +1,6 @@
+oc apply -f hive_v1alpha1_syncset.yaml
```
fahlmant: Is this a testing script as well?

manifests/hive_v1alpha1_syncset.yaml
```diff
@@ -0,0 +1,93 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
fahlmant: Is this file for testing?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -236,16 +237,6 @@ func TestReconcileClusterDeployment(t *testing.T) {
 			setupPDMock: func(r *mockpd.MockClientMockRecorder) {
 			},
 		},
-		{
```
fahlmant: Did you mean to remove this block?

pkg/controller/clusterdeployment/clusterdeployment_created.go
```diff
@@ -21,17 +21,15 @@ import (
...
 	"github.com/openshift/pagerduty-operator/pkg/kube"
+	metrics "github.com/openshift/pagerduty-operator/pkg/metrics"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func (r *ReconcileClusterDeployment) handleCreate(request reconcile.Request, instance *hivev1.ClusterDeployment) (reconcile.Result, error) {
-	if !instance.Status.Installed {
```
fahlmant: Did you mean to remove this block?

pkg/metrics/metrics.go
```diff
@@ -26,13 +30,24 @@ const (
...
+
+	metricPagerDutyHeartbeat = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "metricPagerDutyHeartbeat",
+		Help: "Metric for heartbeating the Dead Man's Snitch api",
 	}, []string{"name"})
 
 	metricsList = []prometheus.Collector{
-		metricPlaceholder,
+		metricPagerDutyCreateFailure,
+		metricPagerDutyDeleteFailure,
```
fahlmant: You will also need to put the heartbeat metric in here

pkg/metrics/metrics.go
```diff
@@ -56,8 +71,39 @@ func RegisterMetrics() error {
 	return nil
 }
 
-// UpdatePlaceholderGauge ...
-func UpdatePlaceholderGauge() {
+// UpdateAPIMetrics updates all API endpoint metrics ever 5 minutes
+func UpdateAPIMetrics() {
```
fahlmant: Is this called anywhere? This will probably need to be run in a go routine

manifests/05-operator.yaml
liamwazherealso: Yes this is temporary for testing purposes.

manifests/apply.sh
liamwazherealso: Yes this is temporary for testing purposes.

manifests/hive_v1alpha1_syncset.yaml
liamwazherealso: Yes this is temporary for testing purposes.

pkg/metrics/metrics.go
liamwazherealso: /hold

liamwazherealso: Ahh, could you point me to an example?

pkg/metrics/metrics.go
liamwazherealso: Updated.

pkg/metrics/metrics.go
fahlmant: https://github.com/openshift/certman-operator/blob/master/cmd/manager/main.go#L120

manifests/hive_v1alpha1_syncset.yaml
liamwazherealso: Removed testing files.

manifests/apply.sh
liamwazherealso: Removed testing files.

liamwazherealso:
> So a new cluster gets deployed, and this operator gets run to setup pagerduty for that new operator.
If it fails, the metric is set to 1. Do we alert when that happens? 
Yes, we will add alerting rules later on. 

> Now say another cluster is created and the reconcile is successful, then the metric is set to 0 and the alert resolves. But the other cluster still didn't provision pagerduty successfully.
In it's current state this would happen. This needs to be addressed.

pkg/metrics/metrics.go
liamwazherealso: Updated

liamwazherealso:
> Question about the metric.
> 
> So a new cluster gets deployed, and this operator gets run to setup pagerduty for that new operator.
> If it fails, the metric is set to 1. Do we alert when that happens? Now say another cluster is created and the reconcile is successful, then the metric is set to 0 and the alert resolves. But the other cluster still didn't provision pagerduty successfully.
> 
> @mwoodson What do we need this operator to report/measure?

I have updated the metrics to use a label that corresponds to the clusterdeployment (clusterid). This we will be able to have different instances of the metric per clusterdeployment. Solving this issue.

liamwazherealso:
@jharrington22 @fahlmant @mwoodson Please review my changes.  Metrics are visible on the route.

pkg/localmetrics/localmetrics.go
```diff
@@ -15,22 +15,72 @@
...
+func UpdateMetricPagerDutyHeartbeatGauge() {
+
+	_, err := http.NewRequest("GET", "api.pagerduty.com", nil)
+	if err != nil {
+		MetricPagerDutyHeartbeat.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(0))
+	}
+
+	pdAPISecret := &corev1.Secret{}
+
+	APIKey := string(pdAPISecret.Data["PAGERDUTY_API_KEY"])
```
fahlmant: Is this API string empty?

pkg/localmetrics/localmetrics.go
```diff
@@ -15,22 +15,81 @@
...
+// UpdateMetricPagerDutyDeleteFailure updates guage to 1 when deletion fails
+func UpdateMetricPagerDutyDeleteFailure(x int, cd string) {
+	MetricPagerDutyDeleteFailure.With(prometheus.Labels{"name": "pagerduty-operator", "clusterdeployment_name": cd}).Set(float64(x))
+}
+
+// UpdateMetricPagerDutyHeartbeatGauge curls the PD API, updates the gauge to 1 when successful.
+func UpdateMetricPagerDutyHeartbeatGauge(APIKey string) {
+
+	_, err := http.NewRequest("GET", "api.pagerduty.com", nil)
+	if err != nil {
```
jharrington22: Should we log an error here? /cc @fahlmant

pkg/localmetrics/localmetrics.go
```diff
@@ -15,22 +15,81 @@
...
+		MetricPagerDutyHeartbeat.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(0))
+	}
 
-	metricPlaceholder.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(1))
+	client := pdApi.NewClient(APIKey)
+
+	if client != nil {
```
jharrington22: Error here too? @fahlmant

pkg/localmetrics/localmetrics.go
fahlmant: +1

pkg/localmetrics/localmetrics.go
fahlmant: +1

pkg/localmetrics/localmetrics.go
liamwazherealso: Fixed

pkg/localmetrics/localmetrics.go
liamwazherealso: Done

pkg/localmetrics/localmetrics.go
liamwazherealso: Done

pkg/localmetrics/localmetrics.go
```diff
@@ -15,22 +15,90 @@
...
+// UpdateMetricPagerDutyDeleteFailure updates guage to 1 when deletion fails
+func UpdateMetricPagerDutyDeleteFailure(x int, cd string) {
+	MetricPagerDutyDeleteFailure.With(prometheus.Labels{"name": "pagerduty-operator", "clusterdeployment_name": cd}).Set(float64(x))
+}
 
-	metricPlaceholder.With(prometheus.Labels{"name": "pagerduty-operator"}).Set(float64(1))
+// UpdateMetricPagerDutyHeartbeatGauge curls the PD API, updates the gauge to 1 when successful.
+func UpdateMetricPagerDutyHeartbeatGauge(APIKey string) {
+	metricLogger := log.WithValues("Namespace", "pagerduty-operator")
+	metricLogger.Info("Reconciling GroupPermission")
```
fahlmant: This string references another operator's CRD :)

liamwazherealso: :)

pkg/localmetrics/localmetrics.go
```diff
@@ -61,33 +63,47 @@ func UpdateAPIMetrics(APIkey string) {
...
+		req.Header.Set("Authorization", "Token token="+APIkey)
+
+		_, err := http.DefaultClient.Do(req)
+
+		if err != nil {
+			metricLogger.Error(err, "Failed to reach api when authenticated")
+			MetricPagerDutyHeartbeat.With(prometheus.Labels{
+				"name": "pagerduty-operator"}).Set(float64(0))
+			return
 		}
```
jharrington22: You want to make sure you close your http requests after you use them otherwise they will not be closed and you'll introduce a resource leak with lots of open http connections that haven't been closed.

You will want to change the request to store the response `_, err` becomes `resp, err` and then add after the err check block `defer resp.Body.Close()`

jharrington22:
@liamwazherealso re test please :)

pkg/localmetrics/localmetrics.go
liamwazherealso: Done

liamwazherealso: Done

liamwazherealso: Done

pkg/localmetrics/localmetrics.go
```diff
@@ -15,22 +15,97 @@
...
+const (
+	apiEndpoint = "https://api.pagerduty.com/users"
 )
 
 var (
-	metricPlaceholder = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "placeholder",
-		Help: "Placeholder",
+	MetricPagerDutyCreateFailure = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "metricPagerDutyCreateFailure",
```
fahlmant: Very small nit, but most of the names are usually in the form of `metric_pagerduty_create_failure` and the Help would be something like `Reports the status of PD service creation`

cmd/manager/main.go
```diff
@@ -124,12 +128,30 @@ func main() {
...
 	// Configure metrics if it errors log the error but continue
+
 	if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
 		log.Error(err, "Failed to configure Metrics")
 	}
 
+	client := mgr.GetClient()
+	pdAPISecret := &corev1.Secret{}
+	err = client.Get(context.TODO(), types.NamespacedName{Namespace: "pagerduty-operator", Name: "pagerduty-api-key"}, pdAPISecret)
+	APIkey := ""
```
jharrington22: Defining this as `var APIKey string` is a better syntax for Go.

liamwazherealso: Done

fahlmant:
@liamwazherealso Please squash, otherwise lgtm

PR #53, opened 2019-08-12
update OWNERS

maorfr:
Since the OWNERS file is a source for the App SRE slack usergroups integraiton, this file has to be updated to prevent warnings in the integration.

Let me know if we should silently ignore such warnings.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,4 +9,3 @@ reviewers:
 - mwoodson
 - wshearn
 - jharrington22
-- blrm


jewzaam:
This is open source.  I don't have a problem with Blair being a reviewer.  @mwoodson ?

jharrington22:
@jewzaam Didn't see your comment. I have no problems with that either just put a hold on it as my comment might get it merged.

PR #54, opened 2019-08-15
Remove API metric

liamwazherealso:
Removal or API metrics for Pagerduty operator. I want to ensure that all our metrics are necessary. I see the API metrics as redundant. Since the failure of creation/deletion would follow a failure of an API call.

https://jira.coreos.com/browse/SREP-1941

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -34,6 +34,7 @@ import (
 
 	routev1 "github.com/openshift/api/route/v1"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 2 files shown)


liamwazherealso:
I can see the case for the authenticated call here, we would get a heads up before PD fails to create/delete a component.

jharrington22:
@liamwazherealso @fahlmant correct me if i'm wrong here but this code actually did two things, one ensure periodically that the operator can reach the PD API and two, ensured that the PD operator can authenticate to the PD API?

I believe that don't need an alert if the PD API is down but I do think its important to know if the API key is valid otherwise as @fahlmant mentioned we'd only find out it can't auth if someone looked at the logs or we realized we weren't getting alerts from a particular. I vote to keep the validation of the API call. cc/ @lisa thoughts?

liamwazherealso:
@jharrington22  You are right, and I do think that this PR should at least remove the unauthenticated API call.

lisa:
@fahlmant @jharrington22 
> I disagree that configuration of the operator is not completely unimportant. If clusters aren't being provisioned correctly because the credentials have expired for example, how would we ever know?

and 

> I believe that we don't need an alert if the PD API is down

Kubernetes supports health checks for running pods, and a chronic inability to communicate with a remote resource is one of those internally held metrics (a la healthz) which can be used to fail the health check. When the liveness and/or readiness (health) checks fail, the pod is terminated and rescheduled. When this continually happens, kube-state-metrics will reveal that the pod is in a crashloop.

Periodically polling the resources a Pod needs (whether that is a database server, or 3rd party API service) and reflecting those health data (again, via healthz) can be very useful to feed into Kubernetes liveness and readiness checks. For me, this is a more desirable way to run the operator because we can see directly from the Pod status why the Pod is crash looping. I would like to see Kubernetes attempt to reschedule the Pod to fix any accessibility issues (eg if a particular Node is partitioned in a strange way).

If the API key is correct during the "hey, are my dependencies available" check then that counts towards the healthiness, otherwise, if the API key is incorrect, it will count as unhealthy. Consider the case where the Secret is fixed ("oops, I forgot to change it from `CHANGEME`"), if the operator is the source of the page, we will need to manually intervene to bounce the pod, whereas if it's part of the health check, Kubernetes will restart it as the Pod is terminated and rescheduled, thus picking up the new value.

cmd/manager/main.go
```diff
@@ -140,20 +135,6 @@ func main() {
 		log.Error(err, "Failed to configure Metrics")
 	}
 
-	client := mgr.GetClient()
-	pdAPISecret := &corev1.Secret{}
-	err = client.Get(context.TODO(), types.NamespacedName{Namespace: "pagerduty-operator", Name: "pagerduty-api-key"}, pdAPISecret)
```
lisa: Is this intended to be deleted (lines 143-152)?

PR #59, opened 2019-09-13
add yaml validation

liamwazherealso:
As a SREP engineer I want to know that the CRDs for a project are valid yaml before deploying. A CRD with bad yaml will break a deploy and also one OLM bundle revision meaning that you cannot "roll forward" by just fixing them yaml and deploying again.

https://jira.coreos.com/browse/SREP-2101

### DIFF
--- a/hack/app_sre_pr_check.sh
+++ b/hack/app_sre_pr_check.sh
@@ -4,6 +4,25 @@
 
 set -exv
 
+CURRENT_DIR=$(dirname "$0")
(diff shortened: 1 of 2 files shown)


hack/app_sre_pr_check.sh
```diff
@@ -4,6 +4,20 @@
 
 set -exv
 
+CURRENT_DIR=$(dirname "$0")
+CRD_DIR="$CURRENT_DIR"/../deploy/crds
+
```
rafael-azevedo: We should add support for also looking in the manifests folder

liamwazherealso: Done

PR #64, opened 2019-10-04
Remove SyncSet if noalerts label is added

jewzaam:
https://jira.coreos.com/browse/SREP-1572

Added SyncSet deletion to handleDelete and reference from a few more places.  It checks the finalizer first and bails if there is no finalizer so should be lightweight.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
@@ -22,10 +22,16 @@ import (
 	metrics "github.com/openshift/pagerduty-operator/pkg/localmetrics"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
 	"github.com/openshift/pagerduty-operator/pkg/utils"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, instance *hivev1.ClusterDeployment) (reconcile.Result, error) {
+	if !utils.HasFinalizer(instance, config.OperatorFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	println("handleDelete")
+
 	ClusterID := instance.Spec.ClusterName
 
 	pdData := &pd.Data{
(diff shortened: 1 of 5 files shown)


pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -22,10 +22,16 @@ import (
...
 	"github.com/openshift/pagerduty-operator/pkg/utils"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, instance *hivev1.ClusterDeployment) (reconcile.Result, error) {
+	if !utils.HasFinalizer(instance, config.OperatorFinalizer) {
+		return reconcile.Result{}, nil
+	}
+	println("handleDelete")
```
wshearn: debug printing

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -37,16 +43,35 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
 		return reconcile.Result{}, err
 	}
 
+	println("1")
```
wshearn: debug printing

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -37,16 +43,35 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
...
+
 	err = pdData.ParseClusterConfig(r.client, request.Namespace, request.Name)
 	if err != nil {
+		println("1 err: " + err.Error())
 		return reconcile.Result{}, err
 	}
 
+	println("2")
```
wshearn: debug printing

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -37,16 +43,35 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
...
+	ssName := request.Name + config.SyncSetPostfix
+	println(ssName)
+	r.reqLogger.Info("Deleting PD SyncSet", "Namespace", request.Namespace, "Name", request.Name)
+	syncset := &hivev1.SyncSet{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: request.Namespace, Name: ssName}, syncset)
+	if err == nil {
+		err = r.client.Delete(context.TODO(), syncset)
+		if err != nil {
+			r.reqLogger.Error(err, "Error deleting SyncSet", "Namespace", request.Namespace, "Name", ssName)
+			return reconcile.Result{}, err
```
wshearn: Couldn't there be a race condition here where the CR is updated after a deletion timestamp?

If it can't delete it wouldn't it be better to let it be until the NS gets cleaned up?

jewzaam: We don't fail if we can't delete the PD service.  I can see letting it fall through for that reason.

PR #65, opened 2019-10-07
Fixed deletion of SyncSet

jewzaam:
It won't be recreated by the SyncSet controller.
It will be deleted if noalerts is applied.
Unit tests updated to check these.
Some refactoring.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -114,7 +114,6 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 
 	// If we don't manage this cluster: log, delete, return
 	if !processCD {
-		r.reqLogger.Info("ClusterDeployment isn't watched by this operator:  " + config.ClusterDeploymentManagedLabel + "=" + instance.GetLabels()[config.ClusterDeploymentManagedLabel] + ", " + config.ClusterDeploymentNoalertsLabel + "=" + instance.GetLabels()[config.ClusterDeploymentNoalertsLabel])
 		return r.handleDelete(request, instance)
 	}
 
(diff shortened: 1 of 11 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -106,43 +106,19 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
...
-		r.reqLogger.Info("Is not a managed cluster")
-		return r.handleDelete(request, instance)
-	}
-
-	// Return if alerts are disabled on the cluster
-	if _, ok := instance.Labels[config.ClusterDeploymentNoalertsLabel]; ok {
-		r.reqLogger.Info("Managed cluster with Alerts disabled", "Namespace", request.Namespace, "Name", request.Name)
+	// If we don't manage this cluster: log, delete, return
+	if !processCD {
+		r.reqLogger.Info("ClusterDeployment isn't watched by this operator:  " + config.ClusterDeploymentManagedLabel + "=" + instance.GetLabels()[config.ClusterDeploymentManagedLabel] + ", " + config.ClusterDeploymentNoalertsLabel + "=" + instance.GetLabels()[config.ClusterDeploymentNoalertsLabel])
```
jharrington22: Do we want to log everytime we see a clusterDeployment that we don't manage? Also I think this line would read easier if we created a new var `proccessCDMsg` (or something) and using `fmt.Sprintf()`

pkg/controller/syncset/syncset_controller.go
```diff
@@ -132,6 +132,14 @@ func (r *ReconcileSyncSet) Reconcile(request reconcile.Request) (reconcile.Resul
 	r.reqLogger = log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	r.reqLogger.Info("Reconciling SyncSet")
 
+	isCDCreated, clusterDeployment, err := utils.CheckClusterDeployment(request, r.client, r.reqLogger)
+
+	// If we don't manage this cluster: log, delete, return
+	if !isCDCreated {
```
jharrington22: Same comment here as with the clusterDeployment logging.

PR #70, opened 2019-10-10
Ignore missing config on delete

jewzaam:
Can't do anything if the config isn't there, it's not a recoverable state.

https://jira.coreos.com/browse/SREP-2250

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_deleted.go
@@ -42,27 +42,32 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
 		BaseDomain: instance.Spec.BaseDomain,
 	}
 	err := pdData.ParsePDConfig(r.client)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
+	// if there is an error, no configuration (configmap or secret) was found.
+	// we are deleting.
+	// if we block deletion because of missing config we will block cluster deletion.
+	// if we do not block we could leave a dangling service in PD.
+	// without the config we cannot fix the dangling service.
+	// THEREFORE: do not fail if we can't find PD config, just continue
+	if err == nil {
+		err = pdData.ParseClusterConfig(r.client, request.Namespace, request.Name)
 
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -42,27 +42,32 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
 		BaseDomain: instance.Spec.BaseDomain,
 	}
 	err := pdData.ParsePDConfig(r.client)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
+	// if there is an error, no configuration (configmap or secret) was found.
```
cblecker: Shouldn't we parse the error to actually confirm this?

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
```diff
@@ -42,27 +42,32 @@ func (r *ReconcileClusterDeployment) handleDelete(request reconcile.Request, ins
...
+	// without the config we cannot fix the dangling service.
+	// THEREFORE: do not fail if we can't find PD config, just continue
+	if err == nil {
+		err = pdData.ParseClusterConfig(r.client, request.Namespace, request.Name)
 
-	err = pdData.ParseClusterConfig(r.client, request.Namespace, request.Name)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
+		if err == nil {
```
cblecker: same here.. shouldn't we parse a specific error condition?

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
rogbas: +1, it would be better to use the `IsNotFound(err)` method from `"k8s.io/apimachinery/pkg/api/errors"`

jharrington22: +1

pkg/controller/clusterdeployment/clusterdeployment_deleted.go
jharrington22: +1

PR #73, opened 2019-10-26
Update SDK and fix cache

mjudeikis:
Update SDK to 0.8.2 and add sync start, wait for sync.

* remove isclean as it makes local development hard.
* bump apis.
* fix cache initialization
* wrap main function for good practices


cmd/manager/main.go
```diff
@@ -128,6 +135,10 @@ func main() {
 		log.Error(err, "")
 		os.Exit(1)
 	}
+	// start cache and wait for sync
+	cache := mgr.GetCache()
+	go cache.Start(stopCh)
+	cache.WaitForCacheSync(stopCh)
```
jharrington22: I'm curious on this, since the cache isn't populated until the controllers are initialized I have used in the past a client from the controller-runtime. This is great, did you find this in the documentation or do some investigating?

cmd/manager/main.go
```diff
@@ -58,6 +58,12 @@ func printVersion() {
 }
 
 func main() {
+	if err := start(); err != nil {
```
jharrington22: Why split this? Don't we error during the start process anyway?

jharrington22:
@mjudeikis looks good, we usually try to put vender/updates in one commit (if they exist) as you have and then squash the rest of the changes into a single commit if you could?

cmd/manager/main.go
mjudeikis: This is a standard pattern in non sdk based controllers. I just knew it exists :) 
What is of interest is that it does not start properly now in v4 cluster as the cache is not started. and it hands for ~30 min for next sync cycle. But I was not able to find what changed

cmd/manager/main.go
mjudeikis: Good practices in go to leave main clean. Just code readability.

mjudeikis:
@jharrington22 done

cmd/manager/main.go
jharrington22: My only issue with this is that we don't do it in our other operators and its different to what is templated from the SDK. If you feel strongly about it I'm happy to leave it otherwise I'd prefer it to match our other operators for consistency.

mjudeikis: I dont feel strong, but its just good practice :) and would be nice to move other too

PR #75, opened 2019-10-31
Detect operator-sdk local up

wanghaoran1988:
This is a same one with https://github.com/openshift/certman-operator/pull/88

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -49,6 +49,10 @@ var (
 	metricsPort = "8080"
 	metricsPath = "/metrics"
 )
+
+// Set const to analyze to determine operator-sdk up local
+const envSDK string = "OSDK_FORCE_RUN_MODE"
+
 var log = logf.Log.WithName("cmd")
 
 func printVersion() {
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -146,10 +150,17 @@ func start() error {
...
+	// detect if operator-sdk up local is run
+	detectSDKLocal := os.Getenv(envSDK)
+
+	// Configure metrics. If it errors, log the error and exit.
+	if detectSDKLocal == "local" {
+		log.Info("Skipping metrics configuration; not running in a cluster.")
+	} else {
+		if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
+			log.Error(err, "Failed to configure Metrics")
+			os.Exit(1)
```
mjudeikis: return err

mjudeikis: and replace top ones too to `returun err`

cmd/manager/main.go
```diff
@@ -146,10 +150,17 @@ func start() error {
...
 
-	// Configure metrics if it errors log the error but continue
-	if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
-		log.Error(err, "Failed to configure Metrics")
-		os.Exit(1)
+	// detect if operator-sdk up local is run
+	detectSDKLocal := os.Getenv(envSDK)
+
+	// Configure metrics. If it errors, log the error and exit.
+	if detectSDKLocal == "local" {
```
mjudeikis: `if os.Getenv(envSDK) == "local"`

wanghaoran1988: @mjudeikis Done

cmd/manager/main.go
wanghaoran1988: @mjudeikis Do you mean replace all the `os.Exit(1)` to `return err` in this `start()` func?

mjudeikis: yes

PR #85, opened 2020-04-24
add more unit tests for the referenced secret

bmeng:
fix some small issues in the code and add more test coverage for the referenced secret.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -133,8 +133,10 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		}
 	}
 
+	scName := config.PagerDutySecretName
 	sc := &corev1.Secret{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: ssName, Namespace: request.Namespace}, sc)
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -133,8 +133,10 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		}
 	}
 
+	scName := config.PagerDutySecretName
```
cblecker, line 136: Do we need to reallocate this or can we just use it directly?

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -508,3 +600,46 @@ func verifyNoConfigMapExists(c client.Client) bool {
...
+	}
+
+	return true
+}
+
+// verifyNoSecretExists verifies that the secret which referred by SyncSet does not exist
+func verifyNoSecretExists(c client.Client, expected *SecretEntry) bool {
+	secretList := &corev1.SecretList{}
+	opts := client.ListOptions{Namespace: testNamespace}
+	err := c.List(context.TODO(), &opts, secretList)
```
cblecker, line 630: Is there a reason we would do a List instead of a Get?

bmeng: Just followed the desigh for the syncset and the configmap which verify the non exists. It is better to keep them the same?

PR #88, opened 2020-05-18
Move to Go modules and Go 1.13

grdryn:


grdryn:
/cc @jewzaam @cblecker 

I guess this might need something akin to https://github.com/openshift/release/pull/9062?

PR #91, opened 2020-06-02
Update make and build stuff to be more consistent

jewzaam:
Taking what is the current state of configure-alertmanager-operator with one small tweak in how CONTAINER_ENGINE is set.
Did not carry forward all scripts as this requires job changes and is a much bigger scope.
Driver was I couldn't build this locally for many previously-solved-in-other-repos reasons.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,94 +1,20 @@
 SHELL := /usr/bin/env bash
 
-# Include project specific values file
-# Requires the following variables:
-# - IMAGE_REGISTRY
-# - IMAGE_REPOSITORY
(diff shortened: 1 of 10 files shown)


jewzaam:
Summary of what I did:
1. copy *.mk and Makefile from configure-alertmanager-operator
1. modify to set `CONTAINER_ENGINE?=docker`
1. copy and edit build/Dockerfile from configure-alertmanager-operator
1. copy and edit build/bin/* from configure-alertmanager-operator
1. edit existing scripts in `hack/` to use `make build` instead of `make docker-build`

PR #96, opened 2020-06-03
Added build/Dockerfile.catalog_registry

jewzaam:


lisa:
Background: This is required because #91 makes reference to this Dockerfile.

/lgtm

PR #97, opened 2020-06-05
Fix yaml validation in PR check

jewzaam:
We haven't (and dont' in master yet) have any yaml to validate so never caught this.  With upcoming work to introduce configuration to the operator the PR check is failing because of this.


grdryn:
/lgtm

That probably won't work from me. :smile: I have rebased #97 on top of this, and the job that was failing beforehand is now passing.

PR #98, opened 2020-06-11
Fix YAML validation

2uasimojo:
- `hack/validate_yaml.py` was a no-op when run under py3 because `yaml.safe_load_all()` produces a generator, and doesn't actually parse the YAML until you iterate on it. Fix.
- Previously `app_sre_pr_check.sh` was assembling YAML directories in an
array, but only "looping" over the first one. Now we intelligently look for all YAML files (.yml or
.yaml suffix) under source control in the current repo and run validation on them one by one.

### DIFF
--- a/hack/app_sre_pr_check.sh
+++ b/hack/app_sre_pr_check.sh
@@ -5,22 +5,16 @@
 set -exv
 
 CURRENT_DIR=$(dirname "$0")
-YAML_DIRS=( "${CURRENT_DIR}/../deploy/crds" "${CURRENT_DIR}/../manifests" )
+REPO_ROOT=$(git rev-parse --show-toplevel)
+YAML_FILES=$(git ls-tree --full-tree -r --name-only HEAD | egrep '\.ya?ml$')
 
-for DIR in $YAML_DIRS 
+for YAML_FILE in $YAML_FILES
 do 
-	if [[ -d $DIR ]]; then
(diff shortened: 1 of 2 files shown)


2uasimojo:
> Yaml still isn't validated in manifests/ because ./hack/app_sre_pr_check.sh for loop needs fixed to correctly iterate over the array.
> 
> From: `for DIR in $YAML_DIRS`
> To: `for DIR in ${YAML_DIRS[@]}`

Mm. I was going to hit the script separately and make it iterate over all YAML files in source control rather than a hardcoded list of subdirectories. But didn't realize it was already busted (I was only testing directly with `validate_yaml.py`). Will fix.

2uasimojo:
Is `app_sre_pr_check.sh` _supposed_ to be exiting before doing `make build`?

jewzaam:
> Is `app_sre_pr_check.sh` _supposed_ to be exiting before doing `make build`?

We have prow building the image so it really doesn't matter probably.  We need to adopt one way of doing pr checks (prow).  But having it run the build should be OK for the ci-ext pr check too.

2uasimojo:
> > Is `app_sre_pr_check.sh` _supposed_ to be exiting before doing `make build`?
> 
> We have prow building the image so it really doesn't matter probably. We need to adopt one way of doing pr checks (prow). But having it run the build should be OK for the ci-ext pr check too.

I think you were saying I could remove the `exit 0`, so I did.

PR #106, opened 2020-07-27
Don't hardcode target Secret details in SyncSet

grdryn:


pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go
```diff
@@ -145,7 +145,8 @@ func testSecret() *corev1.Secret {
 func testSyncSet() *hivev1.SyncSet {
 	secretName := config.Name(testServicePrefix, testClusterName, config.SecretSuffix)
 	secret := kube.GeneratePdSecret(testNamespace, secretName, testIntegrationID)
-	ss := kube.GenerateSyncSet(testNamespace, testClusterName, secret)
+	pdi := testPagerDutyIntegration()
+	ss := kube.GenerateSyncSet(testNamespace, testClusterName, secret, pdi)
```
jewzaam, line 149: We should update unit tests to verify source and target.  Verification is in

https://github.com/openshift/pagerduty-operator/blob/77c5ea5dad8d5e354f1044c7cf77f1b51742c381/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go#L714-L735

and is called here

https://github.com/openshift/pagerduty-operator/blob/77c5ea5dad8d5e354f1044c7cf77f1b51742c381/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go#L501

which would need to add in the PDI reference.

grdryn: I've done it slightly differently than you've suggested: rather than passing in the PDI reference, I added targetSecret as a field on `SyncSetEntry` -- it looks like the purpose of that type is to give info to `verifySyncSetExists` that should be checked. Let me know if you'd like me to change that back to what you've suggested, it's an easy change either way :)

PR #100, opened 2020-07-07
Upgrade to operator-sdk v0.16.2

grdryn:
Followed upgrade guide at:
https://sdk.operatorframework.io/docs/migration/version-upgrade-guide/#v016x

The reason I didn't go all the way to the latest 0.18.z is that after
0.16.z there were dependency conflicts that might best be solved by
either updating dependencies in openshift/operator-custom-metrics, or
by removing that as a dependency from this operator.

PR #107, opened 2020-07-27
Don't use old hardcoded PD API key secret

grdryn:
It should be specified in the PDI CR.

### DIFF
--- a/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
+++ b/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
@@ -52,29 +52,16 @@ var log = logf.Log.WithName("controller_pagerdutyintegration")
 // Add creates a new PagerDutyIntegration Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	newRec, err := newReconciler(mgr)
-	if err != nil {
-		return err
-	}
-
-	return add(mgr, newRec)
+	return add(mgr, newReconciler(mgr))
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
-	tempClient, err := client.New(mgr.GetConfig(), client.Options{Scheme: mgr.GetScheme()})
-	if err != nil {
-		return nil, err
-	}
-
-	// get PD API key from secret
-	pdAPIKey, err := utils.LoadSecretData(tempClient, config.PagerDutyAPISecretName, config.OperatorNamespace, config.PagerDutyAPISecretKey)
-
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	return &ReconcilePagerDutyIntegration{
 		client:   utils.NewClientWithMetricsOrDie(log, mgr, controllerName),
 		scheme:   mgr.GetScheme(),
-		pdclient: pd.NewClient(pdAPIKey),
-	}, nil
+		pdclient: pd.NewClient,
+	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -159,7 +146,7 @@ type ReconcilePagerDutyIntegration struct {
 	client    client.Client
 	scheme    *runtime.Scheme
 	reqLogger logr.Logger
-	pdclient  pd.Client
+	pdclient  func(APIKey string) pd.Client
 }
 
 // Reconcile reads that state of the cluster for a PagerDutyIntegration object and makes changes based on the state read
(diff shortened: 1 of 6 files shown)


pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
```diff
@@ -184,10 +171,21 @@ func (r *ReconcilePagerDutyIntegration) Reconcile(request reconcile.Request) (re
...
 	}
 
+	pdApiKey, err := utils.LoadSecretData(
+		r.client,
+		pdi.Spec.PagerdutyApiKeySecretRef.Name,
+		pdi.Spec.PagerdutyApiKeySecretRef.Namespace,
+		config.PagerDutyAPISecretKey,
+	)
+	if err != nil {
+		return r.requeueOnErr(err)
```
jewzaam, line 181: I worry this will be a hot-loop of requeue.  I don't know if the sdk supports an exponential backoff but it should be used here.  Alternately we do not requeue this.  Reconcile will run periodically and we'll get a retry then.  Note the operator wouldn't have loaded in the prior implementation as the `Add` function would have raised an error.  We shouldn't do that now but we should be resilient to failures and strive for minimal impact on misconfiguration.

Additionally we should collect and export metrics on this failure so we can alert on it.  Misconfigured PDI need to be raised to on-call.

* Metric: pagerdutyinstance_secret_loaded
* Labels:
   * name = the PDI name
* Value:
   * 0 = unable to load referenced secret
   * 1 = able to load referenced secret

Then alert when metric value is `0` for some period of time, say `15m`.

To date we have put SRE alerts into openshiftmanaged-cluster-config.  But it doesn't have to be there.  I'd prefer we include this in the repo in `manifests` and bundle it into `hack/olm-registry/olm-artifacts-template.yaml` for deployment along with the operator.

grdryn: How about a requeue every 10 minutes on failure here?

jewzaam: I think the reconcile loop is run for all resources every 10 minutes already, but I don't have a reference for that.  Having it requeue on some delay would be OK, though we won't know about it w/o metrics and alerts.

grdryn: Since there's no `SyncPeriod` option passed to the manager [here in main.go](https://github.com/openshift/pagerduty-operator/blob/master/cmd/manager/main.go#L107), then [the default is 10 hours](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#Options). I think if we want to set a `RequeueAfter` flag on the reconcile result, then we may need to swallow the error. That may be fine if we want to just log it and have a metric that we can alert on.

grdryn:
@jewzaam I've attempted to add a metric around the ability to load the API key from the secret specified in the CR, but I'm having a little trouble testing it out: I can't see any of the custom metrics showing up on the metrics endpoint (so they don't get to prometheus either). In the operator logs, I can regularly see the log entry for the existing pagerduty API heartbeat metric, but I don't see that show up on the served page at the metrics endpoint. Log:

```json
{"level":"info","ts":1596291862.8925786,"logger":"localmetrics","msg":"Metrics for PD API","Namespace":"pagerduty-operator"}
```

I don't really know enough about the operator-custom-metrics library that's being used here, and don't currently have time to debug (maybe next week). Could you check if those existing custom metrics get served as expected in stage?

2uasimojo:
> I'm having a little trouble testing it out: I can't see any of the custom metrics showing up on the metrics endpoint (so they don't get to prometheus either).

Hi @grdryn.

You've been hit by a recent change in the controller-runtime. In older versions, if you didn't bootstrap the controller-runtime Manager with a MetricsBindAddress, it would **not** switch on the default metrics service. But since [0.2](https://github.com/kubernetes-sigs/controller-runtime/pull/512), which included [this PR](https://github.com/kubernetes-sigs/controller-runtime/pull/510), the default changed to serving up metrics at port `8080`. You accidentally crossed this boundary when you upgraded the operator-sdk via #100.

You normally wouldn't care that controller-runtime enables metrics you don't use. But if you try to register to the same port via operator-custom-metrics, it fails (silently [1]).

Thus, you will have to fix pagerduty-operator so it doesn't encounter this port conflict. The recommended way is to explicitly disable the default metrics service. You can do this by passing `MetricsBindAddress: "0"` to the `Options` you pass to `manager.New()`. Here's an [example](https://github.com/openshift/certman-operator/pull/149/files#diff-9e63f7b0791edcb4a53c944552b88970R95-R99) of such a change.

Hope this helps!

[1] @fahlmant (maintainer of operator-custom-metrics) is aware and is [working on](https://github.com/openshift/operator-custom-metrics/pull/31) getting it to blow up when it encounters this problem.

grdryn:
@2uasimojo thanks for the tip! I've fixed it now and everything seems fine now.

I've also added the pdi name as a label to the existing create/delete metrics.

I've also put in a check to make sure the old finalizer won't block deletion.

cmd/manager/main.go
```diff
@@ -106,6 +106,8 @@ func start() error {
 	// Create a new Cmd to provide shared dependencies and start components
 	mgr, err := manager.New(cfg, manager.Options{
 		Namespace: "",
+		// Disable controller-runtime metrics serving
+		MetricsBindAddress: "0",
```
2uasimojo, line 110: ✔

grdryn:
I have rebased this (conflict with another PR merged earlier today), and it's ready for review again.

PR #108, opened 2020-07-28
Fix data type for resolveTimeout and acknowledgeTimeout in template

jewzaam:
They were being treated as strings by `oc process` and they need to be integers.
Ideally PR check would catch this, not sure how to validate the CR w/o a live cluster though so will open separate PR if/when I can for that.


RiRa12621:
should probably be an e2e test to catch that

PR #111, opened 2020-07-31
README: reword fake ClusterDeployment section

2uasimojo:
This referred to a clusterdeployment YAML manifest that doesn't exist.
Reword to explain grabbing one from a real hive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,11 +145,15 @@ visible in the URL after the `#` character.
 
 `pagerduty-operator` doesn't start reconciling clusters until `spec.installed` is set to `true`.
 
-Create ClusterDeployment.
+You can create a dummy ClusterDeployment by copying a real one from an active hive
 
 ```terminal
+real-hive$ oc get cd -n <namespace> <cdname> -o yaml > /tmp/fake-clusterdeployment.yaml
+
+...
+
 $ oc create namespace fake-cluster-namespace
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,11 +145,15 @@ visible in the URL after the `#` character.
 
 `pagerduty-operator` doesn't start reconciling clusters until `spec.installed` is set to `true`.
 
-Create ClusterDeployment.
+You can create a dummy ClusterDeployment by copying a real one from an active hive
```
jewzaam, line 148: Why snag one from a real cluster if it's not going to provision a cluster?  Seems like you'd be better off with a static CD to reference.  I assume this is just to test the PD things.  In reality you could just pick a random stage cluster to test with, toggling alerts on and off as needed w/ https://github.com/openshift/ops-sop/blob/master/v4/howto/noalerts.md

2uasimojo: > Why snag one from a real cluster if it's not going to provision a cluster?

I don't think you *want* to provision a cluster.

> I assume this is just to test the PD things.

That's what I was assuming, yeah. But I don't actually know.

> In reality you could just pick a random stage cluster to test with

But wouldn't that involve installing your in-dev operator on the real hive stage?

In summary, I thought this section was talking about how to do nondisruptive local testing without a real hive.

PR #115, opened 2020-08-31
Add SKIPPED_LEGALENTITY_IDS array in template

jewzaam:
https://issues.redhat.com/browse/OSD-4817

So we can specify an optional array of legal entity ID's to skip creating PD services for.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -14,6 +14,8 @@ parameters:
   required: true
 - name: ESCALATION_POLICY
   required: true
+- name: ALERTS_SKIP_LEGALENTITY_IDS
+  value: '[]'
 - name: CHANNEL
   value: staging
 - name: IMAGE_TAG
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -14,6 +14,8 @@ parameters:
   required: true
 - name: ESCALATION_POLICY
   required: true
+- name: ALERTS_SKIP_LEGALENTITY_IDS
+  value: '["NOOP"]'
```
jharrington22, line 18: This can't just be a blank array?

jewzaam: probably, don't remember why I did this.  similar in https://github.com/openshift/splunk-forwarder-operator/pull/85

jharrington22: If we can make it blank its less confusing since `NOOP` doesn't mean anything.

jharrington22 (approved):
/lgtm

PR #116, opened 2020-09-01
Increase requests and limits

NautiluX:
* align requests and limits
* increase memory limit

### DIFF
--- a/manifests/05-operator.yaml
+++ b/manifests/05-operator.yaml
@@ -22,11 +22,11 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
-              memory: "100Mi"
-              cpu: "10m"
+              memory: "400Mi"
+              cpu: "100m"
             limits:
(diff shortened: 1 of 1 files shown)


manifests/05-operator.yaml
```diff
@@ -22,10 +22,10 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
-              memory: "100Mi"
-              cpu: "10m"
+              memory: "400Mi"
+              cpu: "30m"
```
jharrington22, line 26: This is a very small amount of CPU. I'm in favor of upping this to 100m

NautiluX: sure, we can do that

jewzaam: We should just remove the CPU bits, this will throttle the operator.  But I'm OK w/ this being in for now given we had it before.

PR #119, opened 2020-09-14
Use correct capitalization for selector operator

cblecker:
Fix the following error:
```
"request":"pagerduty-operator/osd-silent","error":"\"in\" is not a valid pod selector operator"
```


RiRa12621 (approved):
/lgtm

cblecker:
Manually approving, as this is a bug fix in stage that is blocking cluster uninstalls.

PR #124, opened 2020-09-30
Remove code to handle noalerts labels, move to PagerDutyIntegration instances

jewzaam:
Realized this morning we could strip all of this logic out and rely on the clusterDelpoymentSelector for noalerts capabilities.


jewzaam:
Incorporated in https://github.com/openshift/pagerduty-operator/pull/128

PR #128, opened 2020-11-06
Fix to delete PD service for CD that no longer matches PDI

jewzaam:
https://issues.redhat.com/browse/OSD-5594

Unit tests were reworked to cover many more permitations.
Noalerts was removed as part of test rework / simplification.  Therefore this incorporates https://github.com/openshift/pagerduty-operator/pull/124 which will be closed.


fahlmant (approved):
/lgtm

PR #132, opened 2020-12-11
Bump version of hive for new immutable field

jewzaam:


jewzaam:
duplicates https://github.com/openshift/pagerduty-operator/pull/133

PR #137, opened 2021-02-01
OSD-6070 Patch clusterdeployment instead of update

mrbarge:
### What type of PR is this?
Refactor

This PR changes the logic of the `pagerduty-operator` to perform a `Patch()` of the `clusterdeployment` CR when adding or removing finalizers, rather than an `Update()`, in order to eliminate the possibility that the `Update()` call will drop data from the CR.

[OSD-6070](https://issues.redhat.com/browse/OSD-6070)

### Special notes for your reviewer:
I've tested this with a local deployment of `pagerduty-operator` and verified that finalizers are still created and removed as expected. I have also verified that - if the Hive `clusterdeployment` CRD introduces a new field and sets that value in a `clusterdeployment` CR - the currently running operator does not remove that new field when it performs the `Patch()`, as per the goals of [OSD-6070](https://issues.redhat.com/browse/OSD-6070).

Related PRs:
- [DeadmansSnitch](https://github.com/openshift/deadmanssnitch-operator/pull/99)
- [Certman](https://github.com/openshift/certman-operator/pull/175)

- [X] Tested latest changes against a cluster

### DIFF
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
@@ -16,6 +16,7 @@ package pagerdutyintegration
 
 import (
 	"context"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


jewzaam:
/lgtm
/hold

@mrbarge looks good, remove hold when ready to verify in non-prod

PR #138, opened 2021-02-22
Add scale test PagerdutyIntegration

jharrington22:
This PR adds a PagerDutyInegration CR that will select all managed clusters that have a specific organization ID and route them to a new scale test PD escalation policy.

The organization ID will be a specific org that we use for scale tests its added via parameter in the OC template in this [MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/15310)

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -18,6 +18,12 @@ parameters:
   required: true
 - name: SILENT_ALERT_LEGALENTITY_IDS
   value: '["None"]'
+- name: SCALE_TEST_SERVICE_PREFIX
+  required: true
+- name: SCALE_TEST_ESCALATION_POLICY
+  required: true
+- name: SCALE_TEST_LEGALENTITY_IDS
+  value: '["None"]'
 - name: CHANNEL
   value: staging
 - name: IMAGE_TAG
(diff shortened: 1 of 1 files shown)


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -52,6 +58,44 @@ objects:
...
+    clusterDeploymentSelector:
+      matchExpressions:
+      # only create PD service for managed (OSD) clusters
+      - key: api.openshift.com/managed
+        operator: In
+        values: ["true"]
+      # select specific organizations
+      - key: api.openshift.com/legal-entity-id
+        operator: In
+        values: ${{SCALE_TEST_LEGALENTITY_IDS}}
```
jewzaam, line 83: This also needs excluded from the main PDI, else you get 2 secrets managed by hive named the same thing and.. Bad Things.

jewzaam (changes requested):

hack/olm-registry/olm-artifacts-template.yaml
jharrington22: Added the `SCALE_TEST_LEGALENTITY_IDS` selector to each PDI

jharrington22:
I added the `SCALE_TEST_LEGALENTITY_IDS` array instead of an `IGNORE_LEGALENTITIES_IDS` array because the latter wouldn't be specific and a use case could arise where we want to ignore legal entities from all PDIs.

PR #139, opened 2021-02-23
OSD-6375 Handle Hive ClusterDeployment relocation

mrbarge:
### What type of PR is this?

Feature

This PR updates the logic for the `pagerduty-operator` to handle Hive ClusterDeployment relocation events. This introduces the following new logic:

- During a clusterdeployment deletion, the operator will check for the presence of a `relocate` annotation indicating that the clusterdeployment is relocating. If this is the case, the pagerduty.com service will _not_ be deleted, and the SyncSet update type will be changed to `Upsert` prior to deletion so that it does not result in the PD secret being deleted on the remote cluster.

- When reconciling over the Secret, SyncSet and ConfigMap resources for a cluster with an existing PD service set up, the operator will now check for missing `ownerReferences` on these resources (which will occur as part of the relocation to the destination Hive shard). It will re-instate an `ownerReference` if it finds it to be missing.

All other behaviour of the operator is unchanged.

Please see the following for more information on the change: https://docs.google.com/document/d/13J9iiL0wtgBz56B_pGy47hc1uIz5DIC4_K6nw48dEXM/edit

[OSD-6375](https://issues.redhat.com/browse/OSD-6375)
[SDE-806](https://issues.redhat.com/browse/SDE-806)

- [X] Tested latest changes against a cluster

### DIFF
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
@@ -16,7 +16,9 @@ package pagerdutyintegration
 
 import (
 	"context"
+	hiveconstants "github.com/openshift/hive/pkg/constants"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -53,6 +55,15 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 
 	if !cd.Spec.Installed {
 		// Cluster isn't installed yet, return
+		return nil
+	}
+
+	// Flag if the clusterdeployment is in the middle of a Hive relocation
+	if utils.HasAnnotation(cd, hiveconstants.RelocateAnnotation, strings.HasSuffix, string(hivev1.RelocateIncoming), string(hivev1.RelocateOutgoing)) {
+		/*
+			The ClusterDeployment is relocating. We should not attempt to reconcile any
+			related resources whilst this is happening.
+		*/
 		return nil
 	}
 
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
@@ -129,6 +129,21 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 			return err
 		}
 	}
+	// In the event of a ClusterRelocation, owner-references will be lost. Check and re-instate.
+	cm := &corev1.ConfigMap{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: configMapName, Namespace: cd.Namespace}, cm)
+	if err != nil {
+		return err
+	}
+	if len(cm.OwnerReferences) == 0 {
+		if err = controllerutil.SetControllerReference(cd, cm, r.scheme); err != nil {
+			r.reqLogger.Error(err, "Error setting controller reference on configmap")
+			return err
+		}
+		if err := r.client.Update(context.TODO(), cm); err != nil {
+			return err
+		}
+	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
```diff
@@ -143,6 +164,15 @@ func (r *ReconcilePagerDutyIntegration) handleDelete(pdclient pd.Client, pdi *pa
 		r.reqLogger.Error(err, "Error deleting Secret", "Namespace", cd.Namespace, "Name", secretName)
 	}
 
+	// switch the PD syncset to Upsert prior to deletion if the clusterdeployment is relocating
+	if utils.HasAnnotation(cd, hiveconstants.RelocateAnnotation, strings.HasSuffix, string(hivev1.RelocateOutgoing), string(hivev1.RelocateComplete)) {
+		r.reqLogger.Info("Updating PD SyncSet", "Namespace", cd.Namespace, "Name", secretName)
+		err = utils.UpdateSyncSetApplyMode(secretName, cd.Namespace, r.client, hivev1.UpsertResourceApplyMode, r.reqLogger)
```
bmeng, line 170: Why need to change the apply mode here?

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -129,6 +129,21 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: configMapName, Namespace: cd.Namespace}, cm)
+	if err != nil {
+		return err
+	}
+	if len(cm.OwnerReferences) == 0 {
+		if err = controllerutil.SetControllerReference(cd, cm, r.scheme); err != nil {
+			r.reqLogger.Error(err, "Error setting controller reference on configmap")
+			return err
+		}
+		if err := r.client.Update(context.TODO(), cm); err != nil {
```
bmeng, line 143: Update the ownerReferences or recreate the full resource, which one is the better approach do you think?

bmeng (commented):
Another comment, looks like you only handle the outgoing condition during the clusterdeployment_deleted.go.
Do not need do any change for the relocate incoming in clusterdeployment_created.go?

pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
mrbarge: The PD SyncSets use the `Sync` apply mode - the problem with that is that it means that when the SyncSet is deleted by the outgoing operator, Hive will remove the resource it manages on the remote cluster - in this case the PD secret in `openshift-monitoring`. That leaves the cluster in a state where it is potentially not able to report to PD.

So my strategy for relocation has been to change the apply mode to `Upsert` prior to the SyncSet's deletion on the outgoing Hive shard, so that when the syncset is deleted it results in nil impact to the destination cluster.

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
mrbarge: My main reason for `Update()` was again to avoid a situation where the incoming operator would need to delete the SyncSet (which would remove the managed secret on the remote cluster). 

I personally prefer the approach of deleting everything and letting it be recreated from scratch, but.. I also was hoping to avoid any changes at all to remote cluster during the migration process, so that drove me down this path.

mrbarge:
> Another comment, looks like you only handle the outgoing condition during the clusterdeployment_deleted.go.
> Do not need do any change for the relocate incoming in clusterdeployment_created.go?

I was debating having the operator ignore any reconciling during the `created` phase, but in the end didn't think of too many reasons to actually do it. Until Hive copies the `clusterdeployment` over (which happens after everything else has been copied over), the operator won't be trying to reconcile anything. And when the `clusterdeployment` does arrive, all the other resources should be there too. That all said, I'm definitely happy to add that logic in if you think it would be better that way.

PR #140, opened 2021-04-06
Pin to yq 3.4.1 tag

lisa:
See also https://github.com/openshift/osd-metrics-exporter/pull/43


lisa:
@grdryn and @wshearn can you have a look at this and `/approve` if you agree with what's going on?

PR #152, opened 2021-07-27
OSD-7514 Disable Pagerduty for "fake" OSD clusters

ninabauer:
Checks if the clusterdeployment has the annotation "managed.openshift.com/fake" and skips the creation of PD services if true.

### DIFF
--- a/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go
+++ b/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go
@@ -42,18 +42,19 @@ import (
 )
 
 const (
-	testPagerDutyIntegrationName = "testPagerDutyIntegration"
-	testClusterName              = "testCluster"
-	testNamespace                = "testNamespace"
-	testIntegrationID            = "ABC123"
-	testServiceID                = "DEF456"
-	testAPIKey                   = "test-pd-api-key"
-	testEscalationPolicy         = "test-escalation-policy"
-	testResolveTimeout           = 300
-	testAcknowledgeTimeout       = 300
-	testOtherSyncSetPostfix      = "-something-else"
-	testsecretReferencesName     = "pd-secret"
-	testServicePrefix            = "test-service-prefix"
+	testPagerDutyIntegrationName    = "testPagerDutyIntegration"
+	testClusterName                 = "testCluster"
+	testNamespace                   = "testNamespace"
+	testIntegrationID               = "ABC123"
+	testServiceID                   = "DEF456"
+	testAPIKey                      = "test-pd-api-key"
+	testEscalationPolicy            = "test-escalation-policy"
+	testResolveTimeout              = 300
+	testAcknowledgeTimeout          = 300
+	testOtherSyncSetPostfix         = "-something-else"
+	testsecretReferencesName        = "pd-secret"
+	testServicePrefix               = "test-service-prefix"
+	fakeClusterDeploymentAnnotation = "managed.openshift.com/fake"
 )
 
 type SyncSetEntry struct {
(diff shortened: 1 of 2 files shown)


jewzaam:
We have the label based selector.  Why are we not also setting a label so we can do this and not bake anything special into the operator?  This would work for DMS as well.

/hold

sulmer-redhat:
> We have the label based selector. Why are we not also setting a label so we can do this and not bake anything special into the operator? This would work for DMS as well.
> 
> /hold

@jewzaam From what I know, SD team will use the fake cluster annotation when they want to deploy a fake ClusterDeployment. 
I think that they will not label every CD with a "fake" label, they would just create this annotation (Hive also has a "Fake cluster annotation", different from SD one. If we are also asking them to label, then a fake CD will have two "fake" annotation and one "fake" label?)

NautiluX:
I do prefer the code change over the olm-artifacts-template.yaml change for the following reasons:
- The agreed upon specification for fake clusters doesn't need to change to add a label in addition to the annotations (would need to be coordinated with other teams)
- Consistency with how we detect fake clusters in other operators
- Code change is easier to test then config change

pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go
```diff
@@ -167,13 +168,15 @@ func testPagerDutyIntegration() *pagerdutyv1alpha1.PagerDutyIntegration {
 }
 
 // testClusterDeployment returns a fake ClusterDeployment for an installed cluster to use in testing.
-func testClusterDeployment(isInstalled bool, isManaged bool, hasFinalizer bool, isDeleting bool) *hivev1.ClusterDeployment {
+func testClusterDeployment(isInstalled bool, isManaged bool, hasFinalizer bool, isDeleting bool, isFake bool) *hivev1.ClusterDeployment {
```
jharrington22, line 171: Can we add a test for a "fake" ClusterDeployment?

jharrington22 (commented):
@jewzaam both PD and DMS being aware of ClusterDeployments that are "fake" in the same way that Hive identifies a flake ClusterDeployment makes sense to me. Are you suggesting we should also add a label? Do you think we can use that somewhere else where we can't use the annotation?

pkg/controller/pagerdutyintegration/pagerdutyintegration_controller_test.go
sulmer-redhat: I wrote a test for it on line 247:
https://github.com/openshift/pagerduty-operator/pull/152/files#diff-5651ca632758a6747842fd344dc5effcc1cdff0b5892c58727ba750e49752a62R247

jewzaam:
> @jewzaam both PD and DMS being aware of ClusterDeployments that are "fake" in the same way that Hive identifies a flake ClusterDeployment makes sense to me. Are you suggesting we should also add a label? Do you think we can use that somewhere else where we can't use the annotation?

Selectors are limited to using labels hence the request to add a label.  Then this is managed in configuration and isn't in code.  Nothing else would use the label, but use in selectors has been the intent of labels all a long.  I think it should have been a label to start with and this would have been moot but here we are.

NautiluX:
@jewzaam @jharrington22 
DMS operator contains quite some logic to allow selection based on annotations. For the use-case to ignore a CR based on an annotation, which is the only one I see so far, I would rather not duplicate that logic in the PD operator.

cben:
It's easy to set both label and annotation in clusters-service, just make sure it's agreed and documented with Hive. :man_shrugging: 
(IMHO it's easier to make the interface be "always set both" than bothering with deprecating the annotation.)

I think the trivial CS change would only set it on new clusterdeployments, may be a little bit trickier to update existing ones.

PR #161, opened 2021-09-20
OSD-8440 - Fix for orphaned incidents in PagerDuty

bdematte:
When deleting a cluster, we delete incidents before deleting the underlying service in PD and as this is an asynchronous event on PD side, we are waiting to confirm incident closure.
However, in case the incidents don't get closed within the defined maxWait, we continue with the service deletion which lead to orphaned incidents we cannot delete. 
Instead, we should raise an error in case the incidents didn't get closed in time : 
- in case the incidents got eventually deleted, the cleanup will proceed properly with next reconcile
- in case the event got dropped, the operator will send again the closure event which should go through


cblecker:
/cc @NautiluX 

If you have a moment, would you mind reviewing this change?

dofinn:
Will this error generate a metric->alert ? When this situation happens, we will need a PD admin to go in  and manually clean up orhpaned incidents?

PR #163, opened 2021-12-08
detect a fedramp environment and prep a variable for use

yithian:
for [OSD-9237](https://issues.redhat.com/browse/OSD-9237)

### DIFF
--- a/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
+++ b/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
@@ -161,7 +161,7 @@ func (r *ReconcilePagerDutyIntegration) Reconcile(request reconcile.Request) (re
 
 	r.reqLogger = log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	r.reqLogger.Info("Reconciling PagerDutyIntegration")
-	if len(os.Getenv("FEDRAMP")) > 0 {
+	if len(os.Getenv("FEDRAMP")) == 0 {
 		r.reqLogger.Info("FEDRAMP environment variable unset, defaulting to false")
 	} else {
 		r.reqLogger.Info("running in FedRAMP environment: %b", fedramp)
--- a/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
+++ b/pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
@@ -16,6 +16,7 @@ package pagerdutyintegration
 
 import (
 	"context"
+	"os"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
```diff
@@ -159,6 +161,11 @@ func (r *ReconcilePagerDutyIntegration) Reconcile(request reconcile.Request) (re
 
 	r.reqLogger = log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	r.reqLogger.Info("Reconciling PagerDutyIntegration")
+	if len(os.Getenv("FEDRAMP")) > 0 {
```
dustman9000, line 164: should check for zero length
```suggestion
	if len(os.Getenv("FEDRAMP")) == 0 {
```

pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
```diff
@@ -32,16 +33,17 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
 	controllerName = "pagerdutyintegration"
 )
 
+var fedramp = os.Getenv("FEDRAMP") == "true"
```
jewzaam, line 46: Why does this default to "true"? It is inconsistent with the template, which I agree with as we should default to current production behavior.

```suggestion
var fedramp = os.Getenv("FEDRAMP") == "false"
```

jewzaam (changes requested):

pkg/controller/pagerdutyintegration/pagerdutyintegration_controller.go
yithian: i would argue that because unless the environment variable `FEDRAMP` equals `true` exactly, then the variable `fedramp` actually defaults to false. we can reverse the logic though, if you'd like

jewzaam: Depends on how you want to parse the env var, it doesn't have to be the literal string `"true"`, you could support `"True"` and `"TRUE"` but that's an implementation detail.  Default behavior needs to be that it's treated as not(true) when whatever true string is not set.

yithian: if the value of the env variable `FEDRAMP` is `true`, the comparison will evaluate to true, otherwise it will evaluate to false. either way, it will set the value of the go variable `fedramp` to the result of that comparison

i think it's already doing what you want (case-insensitivity aside)... what am i missing?

jewzaam: :facepalm: yup I was reading it wrong.. thanks!

PR #164, opened 2022-01-07
Rely solely on cd.Status.PowerState for new Hive versions fall back to legacy

jharrington22:
We can now rely on the ClusterDeployment Status field PowerState to determine if a cluster is running or not, this field was introduced in openshift/hive#1604

We only check legacy conditions if that field doesn't exist which may be because the running version of Hive does not yet contain support for that field.

We are removing checking legacy Hibernation condition reason Resuming because during that state the cluster is not "Running" and may not check in with Deadmanssnitch causing erroneous alerts.

A hive API version bump was required to support the new cd.Status.PowerState

[OSD-9349](https://issues.redhat.com/browse/OSD-9349)


jharrington22:
cc/ @dofinn @2uasimojo @smunilla

jharrington22:
Fixing tests.

pkg/controller/pagerdutyintegration/clusterdeployment_hibernated.go
```diff
@@ -78,8 +84,30 @@ func (r *ReconcilePagerDutyIntegration) handleHibernation(pdclient pd.Client, pd
 }
 
 func instancesAreRunning(cd *hivev1.ClusterDeployment) bool {
+	// Get hibernation PowerState a new ClusterDeployment Status field indicating if the cluster is running
+	// ie. The cluster is not "Resuming" if the PowerState is "Running", the cluster is operational.
+	// If the field is blank we move on and check the legacy reasons (It may be blank if the running version of
+	// Hive on cluster doesn't yet support it)
+	if cd.Status.PowerState != "" && cd.Status.PowerState == "Running" {
```
2uasimojo, line 91: redundant empty string check

2uasimojo (commented):
Otherwise sane, pending tests

pkg/controller/pagerdutyintegration/clusterdeployment_hibernated.go
jharrington22: yeah sorry, copy paste fumble. Its fixed once I push fixed test!

2uasimojo (commented):
/lgtm

PR #167, opened 2022-02-07
Improve fedramp logging

sam-nguyen7:
Improve fedramp logic by only checking once at operator start up


sam-nguyen7:
Closing in favor of #172

PR #168, opened 2022-02-24
Anonymize all intances of cluster name used in pagerduty operator

luis-falcon:
Change instances where PDO uses cluster name for internal ID


pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -72,7 +73,7 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 		return r.client.Patch(context.TODO(), cd, baseToPatch)
 	}
 
-	ClusterID := cd.Spec.ClusterName
+	ClusterNS := strings.Split(cd.Namespace, "-")
```
fahlmant, line 76:
```suggestion
	clusterNS := strings.Split(cd.Namespace, "-")
	clusterNS = ClusterNS[len(clusterNS)-1]
```

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -93,7 +94,7 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 	}
 
 	pdData := &pd.Data{
-		ClusterID:          cd.Spec.ClusterName,
+		ClusterID:          clusterNS[len(clusterNS)-1],
```
sam-nguyen7, line 97: you brought this up but yeah I think it would be cleaner to have the clusterID as separate var here so we can re-use it if needed.

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -72,7 +73,7 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 		return r.client.Patch(context.TODO(), cd, baseToPatch)
 	}
 
-	ClusterID := cd.Spec.ClusterName
+	clusterNS := strings.Split(cd.Namespace, "-")
```
sam-nguyen7, line 76: do we need to check if the cluster is in fedramp mode before doing this?

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -114,10 +115,10 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 		r.reqLogger.Info("Creating PD service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
 		_, createErr = pdclient.CreateService(pdData)
 		if createErr != nil {
-			localmetrics.UpdateMetricPagerDutyCreateFailure(1, ClusterID, pdi.Name)
+			localmetrics.UpdateMetricPagerDutyCreateFailure(1, pdData.ClusterID, pdi.Name)
```
sam-nguyen7, line 118: clusterID value doesn't change here. Could you give a justification for getting the value out of pdData instead of just using the var here?

pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
```diff
@@ -57,7 +58,7 @@ func (r *ReconcilePagerDutyIntegration) handleDelete(pdclient pd.Client, pdi *pa
 		return nil
 	}
 
-	ClusterID := cd.Spec.ClusterName
+	clusterNS := strings.Split(cd.Namespace, "-")
```
sam-nguyen7, line 61: this is done twice so far. Can we move this to a util package or something that can be re-used?

sam-nguyen7 (changes requested):
Few questions but this is looking good. We'll be using very similar logic to anonymize other operators!

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
fahlmant: I don't see this addressed so I unresolved it

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
luis-falcon: yes makes sense.

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
luis-falcon: basically consistency. I do not like that the functions on top is getting its value from "pdData.ClusterID" and then just right below another functions getting ClusterID. those 2 hold the same value.

PR #169, opened 2022-03-01
OSD-10025: Disable PagerDuty service when a cluster is in limited support

supreeth7:
Resolves: [OSD-10025](https://issues.redhat.com/browse/OSD-10025)
<hr>

- This PR switches the PagerDuty service enabled/disabled based on the `api.openshift.com/limited-support` label in the cluster deployment.
- When `api.openshift.com/limited-support` is set to **true** the cluster related PagerDuty service is disabled and any active alerts are automatically resolved.
- When `api.openshift.com/limited-support` is set to anything else other than true or if the label is deleted the cluster related PagerDuty service is re-enabled and alerting is resumed.

### DIFF
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
@@ -37,28 +37,28 @@ func (r *ReconcilePagerDutyIntegration) handleLimitedSupport(pdclient pd.Client,
 		// Disable PD service and resolve existing service alerts if limited-support label set to true
 		r.reqLogger.Info("The cluster is in limited-support, disabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
 		if err := pdclient.DisableService(pdData); err != nil {
-			r.reqLogger.Error(err, "Error disabling pagerduty service")
+			r.reqLogger.Error(err, "Error disabling PagerDuty service")
 			return err
 		}
 
 		pdData.LimitedSupport = true
 
 		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
-			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
+			r.reqLogger.Error(err, "Error updating PagerDuty cluster config", "Name", configMapName)
 			return err
 		}
 	} else if !hasLimitedSupport && pdData.LimitedSupport {
-		// Enable PD service
+		// Enable PagerDuty service if limited-support label is-not-true/does-not-exist
 		r.reqLogger.Info("The cluster is not in limited-support, enabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
 		if err := pdclient.EnableService(pdData); err != nil {
-			r.reqLogger.Error(err, "Error enabling pagerduty service")
+			r.reqLogger.Error(err, "Error enabling PagerDuty service")
 			return err
 		}
 
 		pdData.LimitedSupport = false
 
 		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
-			r.reqLogger.Error(err, "Error updating pd cluster config", "Name", configMapName)
+			r.reqLogger.Error(err, "Error updating PagerDuty cluster config", "Name", configMapName)
 			return err
 		}
 	}
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
@@ -39,7 +39,7 @@ func (r *ReconcilePagerDutyIntegration) handleLimitedSupport(pdclient pd.Client,
 			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
 			return err
 		}
-	} else if ok && hasLimitedSupport == "false" {
+	} else if ok {
 		// Re-enable the pagerduty service if limited-support label set to false
 		r.reqLogger.Info("Limited-support set to false, enabling PD service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
 		if err := pdclient.EnableService(pdData); err != nil {
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
@@ -33,7 +33,7 @@ func (r *ReconcilePagerDutyIntegration) handleLimitedSupport(pdclient pd.Client,
 		hasLimitedSupport = val
 	}
 
-	if hasLimitedSupport && !pdData.ServiceDisabled {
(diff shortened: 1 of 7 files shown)


pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,56 @@
...
+		r.reqLogger.Info("Limited-support set to true, disabling pagerduty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.DisableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error disabling pagerduty service")
+			return err
+		}
+		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
+			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
+			return err
+		}
+	} else if ok && hasLimitedSupport == "false" {
```
jewzaam, line 42:
```suggestion
	} else if ok {
```

If the label is not set it should be treated as not(limited-support) always.  If the value is anything other than `"true"` also treat as not(limited-support).

jewzaam (changes requested):
/hold

Open question generally in JIRA on why this is in code VS using PDI selector.  If that conversation is resolved and we want to do this disable then please remove hold (aka don't wait for me).

jewzaam (changes requested):
Conclusion from conversations is we can update the "osd" PDI and delete the secret when limited support label is present.  Addons do not depend on the "osd" PDI's PD secret.

Please update to move this into the "osd" PDI configuration.  Thanks!

supreeth7:
Thank you for looking into this and figuring out the ideal approach!
I will be closing this PR in favor of #175.

supreeth7:
After the recent discussions, conclusion; PD service history matters to SRE hence deleting the service isn't a good idea. It is ideal to go with this (enable/disable PD service) approach instead. And it further ensures consistency with the existing hibernation flow.

pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,56 @@
...
+		r.reqLogger.Info("Limited-support set to true, disabling pagerduty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.DisableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error disabling pagerduty service")
+			return err
+		}
+		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
+			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
+			return err
+		}
+	} else if ok {
```
wshearn, line 42: What happens if the label is deleted instead of being set to false? Wouldn't the cluster stay in limited support?

supreeth7: @wshearn thank you for pointing this out. I assumed the label will always exist with a value, this has now been fixed in a way that prevents redundant calls to PD API to check if a service has been disabled each the label doesn't exist.

supreeth7:
@jewzaam can you please give this a one final look and approve if everything looks good?

pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,67 @@
...
+		r.reqLogger.Info("The cluster is not in limited-support, enabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.EnableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error enabling pagerduty service")
+			return err
+		}
+
+		pdData.LimitedSupport = false
+
+		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
+			r.reqLogger.Error(err, "Error updating pd cluster config", "Name", configMapName)
```
jewzaam, line 61:
```suggestion
			r.reqLogger.Error(err, "Error updating PagerDuty cluster config", "Name", configMapName)
```

pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,67 @@
...
+
+		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
+			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
+			return err
+		}
+	} else if !hasLimitedSupport && pdData.LimitedSupport {
+		// Enable PD service
+		r.reqLogger.Info("The cluster is not in limited-support, enabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.EnableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error enabling pagerduty service")
```
jewzaam, line 54:
```suggestion
			r.reqLogger.Error(err, "Error enabling PagerDuty service")
```

pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,67 @@
...
+		r.reqLogger.Info("The cluster is in limited-support, disabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.DisableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error disabling pagerduty service")
+			return err
+		}
+
+		pdData.LimitedSupport = true
+
+		if err := pdData.SetClusterConfig(r.client, cd.Namespace, configMapName); err != nil {
+			r.reqLogger.Error(err, "Error updating cluster config", "Name", configMapName)
```
jewzaam, line 47:
```suggestion
			r.reqLogger.Error(err, "Error updating PagerDuty cluster config", "Name", configMapName)
```

pkg/controller/pagerdutyintegration/clusterdeployment_limited_support.go
```diff
@@ -0,0 +1,67 @@
...
+	hasLimitedSupport := false
+	if val, err := strconv.ParseBool(cd.Labels[config.ClusterDeploymentLimitedSupportLabel]); err == nil {
+		hasLimitedSupport = val
+	}
+
+	if hasLimitedSupport && !pdData.LimitedSupport {
+		// Disable PD service and resolve existing service alerts if limited-support label set to true
+		r.reqLogger.Info("The cluster is in limited-support, disabling PagerDuty service", "ClusterID", pdData.ClusterID, "BaseDomain", pdData.BaseDomain)
+		if err := pdclient.DisableService(pdData); err != nil {
+			r.reqLogger.Error(err, "Error disabling pagerduty service")
```
jewzaam, line 40:
```suggestion
			r.reqLogger.Error(err, "Error disabling PagerDuty service")
```

jewzaam (commented):
/approve

Small nits on error logs, not a blocking thing but I do like consistency.

PR #172, opened 2022-03-03
Changing instances of cluster name in PDO for cluster ID in fedramp e…

luis-falcon:
Changing instances of cluster name in PDO for cluster ID in fedramp environments

### DIFF
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_created.go
@@ -72,7 +72,10 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 		return r.client.Patch(context.TODO(), cd, baseToPatch)
 	}
 
-	ClusterID := cd.Spec.ClusterName
+	clusterID := cd.Spec.ClusterName
+	if config.IsFedramp() {
+		clusterID = utils.GetClusterID(cd.Namespace)
+	}
 
 	pdAPISecret := &corev1.Secret{}
 	err := r.client.Get(
@@ -93,7 +96,7 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 	}
 
 	pdData := &pd.Data{
-		ClusterID:          cd.Spec.ClusterName,
+		ClusterID:          clusterID,
 		BaseDomain:         cd.Spec.BaseDomain,
 		EscalationPolicyID: pdi.Spec.EscalationPolicy,
 		AutoResolveTimeout: pdi.Spec.ResolveTimeout,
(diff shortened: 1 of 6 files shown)


jewzaam (changes requested):
Why not simply use the actual ID instead of name for all environments?  What is the value of having it be different?  The property set in the PD data is "ClusterID" so I'm surprised we use name.  Name is not unique and we shouldn't care about it.  The thing we care about is ID...

sam-nguyen7:
> Why not simply use the actual ID instead of name for all environments? What is the value of having it be different? The property set in the PD data is "ClusterID" so I'm surprised we use name. Name is not unique and we shouldn't care about it. The thing we care about is ID...

@jewzaam I don't have a clear answer on this but I know that this was briefly discussed at the leads level and it was decided for only fedramp. cc @blrm 
https://coreos.slack.com/archives/C01M1NRKUF4/p1646142273310629?thread_ts=1646081822.790079&cid=C01M1NRKUF4

jewzaam:
> > Why not simply use the actual ID instead of name for all environments? What is the value of having it be different? The property set in the PD data is "ClusterID" so I'm surprised we use name. Name is not unique and we shouldn't care about it. The thing we care about is ID...
> 
> @jewzaam I don't have a clear answer on this but I know that this was briefly discussed at the leads level and it was decided for only fedramp. cc @blrm https://coreos.slack.com/archives/C01M1NRKUF4/p1646142273310629?thread_ts=1646081822.790079&cid=C01M1NRKUF4

Understood the ask is for fedramp only.  But there is a misalignment in the parameter to PD and the value passed.  And we should not care about the cluster name.  I don't see value in having this logic specific to fedramp unless someone can state why cluster name needs to remain for non-fedramp.

@blrm @luis-falcon thoughts?

fahlmant:
Just curious. what does the resulting PD service look like? Do you have a screenshot or something to share?

luis-falcon:
@jewzaam  The first changes I made had non of this logic. Just a simple replacement using the clusterID instead of the name for every environment. I am of the same opinion as you that just keeping the name for non fedramp and id for fedramp is just extra steps for nothing of real impact really. I  also do not see the need for the cluster name if we had the ID or to make a distinction  but I had discuss this with blair previously but apparently in lead talks this separation was decided so i just coded it this way. Would be nice to have the reasoning though :)

blrm:
I do agree that we should ultimately just move to cluster ID for everything. The reasoning for keeping this change scoped to only fedramp for now is explained in the JIRA card here: https://issues.redhat.com/browse/OSD-9561?focusedCommentId=19589360&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19589360

Given the limited time we have to deliver the fedramp environment and the blast radius of swapping out all existing PD services for their Cluster ID replacements, we decided to just target fedramp initially and then move to cluster IDs for the whole fleet in a subsequent epic.

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
```diff
@@ -73,6 +73,9 @@ func (r *ReconcilePagerDutyIntegration) handleCreate(pdclient pd.Client, pdi *pa
 	}
 
 	ClusterID := cd.Spec.ClusterName
```
sam-nguyen7, line 75:
```suggestion
	clusterID := cd.Spec.ClusterName
```

This variable shouldn't be exported.

pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
```diff
@@ -58,6 +58,9 @@ func (r *ReconcilePagerDutyIntegration) handleDelete(pdclient pd.Client, pdi *pa
 	}
 
 	ClusterID := cd.Spec.ClusterName
```
sam-nguyen7, line 60:
```suggestion
	clusterID := cd.Spec.ClusterName
```
This variable shouldn't be exported.

sam-nguyen7 (changes requested):

jewzaam:
Following up in JIRA, thanks for the link and replies.

pkg/controller/pagerdutyintegration/clusterdeployment_created.go
luis-falcon: good catch sam. thanks

sam-nguyen7 (commented):
/lgtm
@luis-falcon any troubles with testing this before we merge?

sam-nguyen7 (approved):

blrm:
@luis-falcon  If testing for this one went well, this looks good to me. @jewzaam Could you PTAL? Thanks!

PR #175, opened 2022-03-05
Update PDI to not alert for 'limited-support' clusters

supreeth7:
Resolves: [OSD-10025](https://issues.redhat.com/browse/OSD-10025)
<hr>

This PR updates the PagerDuty integration template to ensure that a PD service is deleted if the cluster is in limited-support. (aka do not page SRE)
More context in the JIRA.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -143,6 +143,10 @@ objects:
       - key: ext-managed.openshift.io/noalerts
         operator: NotIn
         values: ["true"]
+      # ignore CD w/ limited-support label
+      - key: api.openshift.com/limited-support
+        operator: NotIn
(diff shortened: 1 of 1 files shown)


supreeth7:
@jewzaam can I please have your eyes on this, thanks.

karthikperu7:
Discussing this with @fahlmant on slack, should we keep this consistent with what happens when a cluster is hibernated as in both cases, the cluster can come back and the alert history is valuable? https://github.com/openshift/pagerduty-operator/blob/4ebfdc0ec5547bce765d8b1f2c70d4206f17181b/pkg/controller/pagerdutyintegration/clusterdeployment_hibernated.go#L61

supreeth7:
As discussed in slack, we are going forward with enable/disable approach instead. Hence closing this.

PR #178, opened 2022-03-29
Make controller PDI escalation policy aware

supreeth7:
### **What type of PR is this?**
Feature

- Addons team now allow multiple on-boarding paths which may require handover between the product/service team and SRE.
- In the process, when the PDI escalation policy is updated, it is expected that the EscalationPolicy for existing services are updated as well. PagerDuty operator currently lacks this feature.
- In such case,  the addons team updated the `matchLabel` value in the PDI to an arbitrary string which would not match any existing clusters. This ensured the existing secrets deployed to clusters with RHODS installed were deleted as well as the services in PagerDuty.  They then reverted that change so that the ClusterDeployments re-entered scope for the pagerduty-operator causing both the services and secrets to be recreated with the updated EscalationPolicy
- This PR ensures dynamic reconciliation of existing PagerDuty service's EscalationPolicy against PDI CR's EscalationPolicy.

Resolves: [OSD-10475](https://issues.redhat.com/browse/OSD-10475)

- [x] Tested the changes against a stage cluster.


pkg/pagerduty/service.go
```diff
@@ -337,6 +347,36 @@ func (c *SvcClient) DisableService(data *Data) error {
...
+	}
+
+	service.EscalationPolicy.ID = escalationPolicy.ID
+
+	_, err = c.PdClient.UpdateService(*service)
+	if err != nil {
+		return err
+	}
+
+	if err := c.resolvePendingIncidents(data); err != nil {
```
Ajpantuso, line 369: Is it necessary to resolve open incidents in this scenario? My understanding is that _PagerDuty_ binds an _Incident_ to the _Escalation Policy_ which was present at the time of _Incident_ creation which is preferred in this case. The assignee for unresolved _Incidents_ generated prior to the change should probably manually reassign if they feel they are no longer responsible.

sandhyadalavi: in some cases, it can override manual incidents assignments, too!

Ajpantuso: That is correct. For this feature I don't think we want to impact open incidents. We can request a new feature at a later time if we desire such functionallity.

supreeth7: Thank you for confirming, makes sense. I've pushed the changes.

Ajpantuso (approved):
/lgtm

PR #183, opened 2022-04-03
[OSD-9561] Scrub PD service names for FedRamp clusters

blrm:
This resolves an issue with the changes made to anonymize pagerduty services for the story https://issues.redhat.com/browse/OSD-9561. This updates the `service` package to check for the presence of a `FEDRAMP` environment variable. If present and true, the PD service name is set to the internal cluster ID and the description will be left blank. If `FEDRAMP` is false, the pre-existing service names and descriptions are used.

### DIFF
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -229,8 +229,8 @@ func (c *SvcClient) CreateService(data *Data) (string, error) {
 	}
 
 	clusterService := pdApi.Service{
-		Name:                   data.ServicePrefix + "-" + data.ClusterID + "." + data.BaseDomain + "-hive-cluster",
-		Description:            data.ClusterID + " - A managed hive created cluster",
+		Name:                   generatePDServiceName(data),
+		Description:            generatePDServiceDescription(data),
 		EscalationPolicy:       *escalationPolicy,
 		AutoResolveTimeout:     &data.AutoResolveTimeout,
 		AcknowledgementTimeout: &data.AcknowledgeTimeOut,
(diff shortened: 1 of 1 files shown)


sam-nguyen7 (approved):
/lgtm

pkg/pagerduty/service.go
```diff
@@ -453,6 +453,26 @@ func parseIncidentNumbers(incidents []pdApi.Incident) []uint {
 	return incidentNumbers
 }
 
+// generateServiceName checks if FedRamp is enabled. If it is, it returns
+// an anonymized PD service name.
+func generatePDServiceName(data *Data) string {
+	if config.IsFedramp() {
+		return data.ClusterID
```
jewzaam, line 460:
```suggestion
		return data.ServicePrefix + "-" + data.ClusterID
```

Any reason not to prefix?  If we don't do this the name is not unique if we have more than one PDI associated with a cluster.

pkg/pagerduty/service.go
```diff
@@ -453,6 +453,26 @@ func parseIncidentNumbers(incidents []pdApi.Incident) []uint {
...
+	} else {
+		return data.ServicePrefix + "-" + data.ClusterID + "." + data.BaseDomain + "-hive-cluster"
+	}
+}
+
+// generateServiceDescription checks if FedRamp is enabled. If it is, it returns
+// an empty PD service description
+func generatePDServiceDescription(data *Data) string {
+	if config.IsFedramp() {
+		return ""
```
jewzaam, line 470: Why leave this blank?

jewzaam (changes requested):

pkg/pagerduty/service.go
blrm: The general guidance for FedRAMP alerting has been to provide as little information as possible in these alerts while still providing a way to identify the cluster so that we can access the alertmanager instance directly for additional information. But the possibility of multiple PDIs hadn't occurred to me. I'll add back the prefix.

pkg/pagerduty/service.go
blrm: As mentioned in my post above, I removed this strictly to minimize the amount of information shared to an external service for a FedRAMP cluster.

PR #184, opened 2022-04-04
[OSD-10751] Fix bug where the operator was failing to delete some alerts

mjlshen:
This bug occurs sometimes when a PagerDuty service has more than one integration because `GetIntegrationKey` returns only one integration, while `resolveAlert` (renamed from `resolveIncident`) needs a correct integration key + alert key combination to successfully resolve an alert.

This was not too obvious because the PagerDuty Events V2 API (which `resolveAlert` uses) returns an HTTP 202 (success) regardless if the integration key + alert key combination matches, the API simply acknowledges that the event was received and contains all the necessary arguments. Luckily, alerts themselves indicate the integration that generated them, so the fix wasn't too bad (can isolate onto the second commit for the entire fix). I validated this fix locally by identifying clusters were failing to have alerts resolve (all had two integrations), then tried to resolve them via the Events V2 API locally (https://developer.pagerduty.com/api-reference/b3A6Mjc0ODI2Nw-send-an-event-to-pager-duty) with the "other" integration key with the specified alert and was able to see the alert resolve.

However, it was quite difficult for me to build up enough knowledge to get here due to the lack of "useful" unit tests, so I built them up to a point where I finally understood enough to identify the problem. I don't have a unit test for the fixed function itself yet as I would need to mock the stateful action of resolving an alert, which can be done, but I figured that since this is an ongoing problem that it should be resolved first and the tests can continue to be fleshed out.

### DIFF
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -369,28 +369,26 @@ func (c *SvcClient) UpdateEscalationPolicy(data *Data) error {
 
 // resolvePendingIncidents loops over all unresolved incidents to resolve all contained alerts
 func (c *SvcClient) resolvePendingIncidents(data *Data) error {
-
 	incidents, err := c.getIncidents(data)
 	if err != nil {
 		return err
 	}
 
-	if len(incidents) > 0 {
-		serviceKey, err := c.GetIntegrationKey(data)
+	for _, incident := range incidents {
+		alerts, err := c.PdClient.ListIncidentAlerts(incident.Id)
 		if err != nil {
 			return err
 		}
 
-		for _, incident := range incidents {
-			alerts, err := c.PdClient.ListIncidentAlerts(incident.Id)
+		for _, alert := range alerts.Alerts {
+			integration, err := c.PdClient.GetIntegration(data.ServiceID, alert.Integration.ID, pdApi.GetIntegrationOptions{})
 			if err != nil {
 				return err
 			}
-			for _, alert := range alerts.Alerts {
-				err = c.resolveAlert(serviceKey, alert.AlertKey)
-				if err != nil {
-					return err
-				}
+
+			err = c.resolveAlert(integration.IntegrationKey, alert.AlertKey)
+			if err != nil {
+				return err
 			}
 		}
 	}
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -146,6 +146,8 @@ type Data struct {
 }
 
 // ParseClusterConfig parses the cluster specific config map and stores the IDs in the data struct
+// SERVICE_ID, INTEGRATION_ID, and ESCALATION_POLICY_ID are required ConfigMap data fields, while
+// HIBERNATING and LIMITED_SUPPORT are optional.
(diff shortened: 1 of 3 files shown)


pkg/pagerduty/service.go
```diff
@@ -365,39 +367,35 @@ func (c *SvcClient) UpdateEscalationPolicy(data *Data) error {
...
+	for _, incident := range incidents {
+		alerts, err := c.PdClient.ListIncidentAlerts(incident.Id)
 		if err != nil {
 			return err
 		}
 
-		for _, incident := range incidents {
-			alerts, err := c.PdClient.ListIncidentAlerts(incident.Id)
+		for _, alert := range alerts.Alerts {
+			err = c.resolveAlert(alert.Integration.ID, alert.AlertKey)
```
NautiluX, line 384: Are you sure `alert.Integration.ID` is the right value to pass in to the events API as routing key?
From reading the [Events API reference](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODI2Nw-send-an-event-to-pager-duty), it looks like the routing key is the integration key:
> The GUID of one of your Events API V2 integrations. This is the "Integration Key" listed on the Events API V2 integration's detail page.
Example value (see example on the right side):
```
  "routing_key": "e93facc04764012d7bfb002500d5d1a6",
```

while the Integration.ID seems to be the ID of an integration, comparing above example value to the example from the alerts [API reference](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE0Mw-list-alerts-for-an-incident):
```
      "integration": {
        "id": "PQ12345",
```
Would we probably need to transform it into a integration key like done [here](https://github.com/openshift/pagerduty-operator/blob/01dfafecc82a7713dac35f103caa2702a90eb687/pkg/pagerduty/service.go#L212)?

mjlshen: Yes definitely, good catch! I even renamed the variable from `serviceKey` to `integrationKey`, but still made this mistake haha

pkg/pagerduty/service.go
```diff
@@ -469,15 +472,26 @@ func generatePDServiceDescription(data *Data) string {
...
+		Action:     "resolve",
+		DedupKey:   alertKey,
+		Payload: &pdApi.V2Payload{
+			Summary:  "Cluster does not exist anymore",
+			Source:   "pagerduty-operator",
+			Severity: "info",
+		},
+	}
+
+	// TODO: If the response status is 429 (TooManyRequests) we should probably wait for longer
```
jewzaam, line 491: Requeue and try again later.  Don't block and wait in the controller.

jewzaam (approved):
Seems reasonable.  I will approve.  One question just to confirm.  When all alerts in an incident are resolved, the incident itself auto-resolves?

pkg/pagerduty/service.go
mjlshen: Yes, as soon as all alerts in an incident are resolved, the incident resolves.

Totally agreed on the second point too! I left the TODO for a planned `requeueAfter` some delay if PagerDuty is throttling us.

PR #185, opened 2022-04-05
[OSD-10751] Catching up mock/unit testing to the fix

mjlshen:
#184 contained the fix for this ticket, but went in without getting the unit tests fully fleshed out enough to actually test the change. This PR extends unit tests to reach up to those changes.

* While testing, I found useful helper functions in go-pagerduty v1.4.0, so I bumped us up from v1.2.0 (latest is v1.5.0)
    * Allowed for the removal of `SvcClient.ManageEvent`/`ManageEventFunc`
    * Allowed the use of `ListIncidentAlertsWithOpts`
* The operator was processing all incidents and contained alerts instead of just "unresolved" ones
* Added tests for `GetUnresolvedIncidents`, `GetUnresolvedAlerts`, and `ResolvePendingIncidents`

### DIFF
--- a/pkg/pagerduty/service_mock.go
+++ b/pkg/pagerduty/service_mock.go
@@ -14,14 +14,159 @@ import (
 )
 
 const (
-	mockEscalationPolicyId string = "ESC1"
-	mockIncidentId         string = "INC1"
-	mockIntegrationId      string = "INT1"
-	mockServiceId          string = "SVC1"
+	mockAlertId             string = "ALT1"
+	mockAlertId2            string = "ALT2"
+	mockEscalationPolicyId  string = "ESC1"
+	mockEscalationPolicyId2 string = "ESC2"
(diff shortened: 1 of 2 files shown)


pkg/pagerduty/service.go
```diff
@@ -409,11 +410,25 @@ func (c *SvcClient) getIncidents(data *Data) ([]pdApi.Incident, error) {
 	return incidentsRes.Incidents, err
 }
 
+// getUnresolvedAlerts returns a slice of unresolved incidents for the provided Service ID
+func (c *SvcClient) getUnresolvedAlerts(incidentId string) ([]pdApi.IncidentAlert, error) {
```
dustman9000, line 414: I only see the function reference in the service_test, was that intentional?

dustman9000:
Changes look good to me. I'll approve this, thanks!
/approve
/lgtm

PR #186, opened 2022-04-06
[OSD-10189] Accumulate errors during the reconcile loop instead of immediately bailing

mjlshen:
The main idea is to accumulate errors in a new `type pdiReconcileErrors []error` which implements the builtin error interface, allowing a full reconcile loop to complete before requeuing.

This is important because the operator effectively tries to reconcile all ClusterDeployments each time, attempting delete actions before any creates. This meant that any bugs in PagerDuty service deletion would cause the operator to be unable to attempt to create new PagerDuty services until the service deletion was remediated.

Along the way I found `utils.LoadSecretData` was duplicated by `pagerduty.GetSecretKey`, so I removed GetSecretKey. Likewise, `utils. GetHashOfPullSecret` was unused and so removed.

### DIFF
--- a/pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
+++ b/pkg/controller/pagerdutyintegration/clusterdeployment_deleted.go
@@ -19,8 +19,6 @@ import (
 	"fmt"
 	"strings"
 
-	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


mjlshen:
The tests will need a significant refactor, but wanted to propose this as a solution

PR #187, opened 2022-04-07
[OSD-10917] Default escalationpolicyid to prevent errors in deletion

mjlshen:
Created a new function, `NewData`, to default the "Data" struct to contain PDIEscalationPolicyID by default.

Then, updated `ParseClusterConfig` such that EscalationPolicyID defaults to PDIEscalationPolicyID when missing.

### DIFF
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -25,6 +25,7 @@ import (
 
 	pdApi "github.com/PagerDuty/go-pagerduty"
 	"github.com/openshift/pagerduty-operator/config"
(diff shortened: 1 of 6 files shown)


pkg/pagerduty/service.go
```diff
@@ -143,8 +144,22 @@ type Data struct {
 	LimitedSupport     bool
 }
 
+func NewData(pdi *pagerdutyv1alpha1.PagerDutyIntegration) (*Data, error) {
```
dustman9000, line 147: Add function comments to describe it?

PR #197, opened 2022-07-18
OSD-12437 Use upstream `UpdateService`

mjlshen:
`UpdateService()` previously had a bug so that we needed to implement our own workaround, however it has since been fixed in https://github.com/PagerDuty/go-pagerduty/pull/448, so removing our custom wrapper since we no longer need it!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,10 +4,11 @@ go 1.13
 
 require (
 	cloud.google.com/go v0.57.0 // indirect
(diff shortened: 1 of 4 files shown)


jewzaam:
/hold

1.5.1 was cut in April but the fix was merged 7 days ago.  So this fix isn't in the version updated in this PR, right?

jewzaam:
/hold cancel

commit w/ the fix is what this references, I missed that.  carry on :)

PR #202, opened 2022-08-17
OSD-12845 Remove any configured alert grouping to workaround upstream bug

mjlshen:
In [OSD-12845](https://issues.redhat.com//browse/OSD-12845), we found that PDO was failing to disable a PagerDuty service after the cluster was placed into limited support. This boiled down to an upstream bug when updating a PagerDuty service has time-based Alert Grouping enabled: https://github.com/PagerDuty/go-pagerduty/issues/458 where calls to `UpdateService` result in the following error due to a JSON unmarshalling bug:

```
HTTP response failed with status code 400,
message: Invalid Input Provided (code: 2001): Config is invalid.
```

Since time-based Alert Grouping is something RL/FL's can do at their discretion and it's quite normal for such services to be eventually put into limited support, this is can be considered a bandaid fix that unsets all alert grouping when enabling/disabling a service. 

However, I did also think it made sense to remove Alert Grouping when enabling/disabling services as the service is expected to either be either:
* fully supported after being enabled from hibernation/limited support (and not need Alert Grouping I think?) or
* disabled, in which case we don't care about Alert Grouping at all

Thanks @mrbarge for the bug report!

### DIFF
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -44,7 +44,7 @@ func getConfigMapKey(data map[string]string, key string) (string, error) {
 	return retString, nil
 }
 
-//Client is a wrapper interface for the SvcClient to allow for easier testing
+// Client is a wrapper interface for the SvcClient to allow for easier testing
 type Client interface {
 	GetService(data *Data) (*pdApi.Service, error)
 	GetIntegrationKey(data *Data) (string, error)
@@ -71,7 +71,7 @@ type PdClient interface {
 
 type DelayFunc func(time.Duration)
 
-//SvcClient wraps pdApi.Client
+// SvcClient wraps pdApi.Client
 type SvcClient struct {
 	APIKey   string
 	PdClient PdClient
@@ -324,6 +324,8 @@ func (c *SvcClient) EnableService(data *Data) error {
 		return err
 	}
 
+	service = removeAlertGrouping(service)
+
 	if service.Status != "active" {
 		service.Status = "active"
 		_, err = c.PdClient.UpdateService(*service)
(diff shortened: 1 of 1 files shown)


jewzaam:
/hold

1. before removing the alert grouping on enable please verify this is expected behavior
2. what is the timeline for PD to fix this upstream and make it available for us to consume instead of working around it?

mjlshen:
1. This isn't expected behavior, you are allowed to disable a PagerDuty service, but keep alert grouping functionality. I verified this by just using curl to disable this service so that PDO would stop continuously failing on the specific affected service.
2. It's is a JSON unmarshalling bug and I made an upstream PR to fix it: https://github.com/PagerDuty/go-pagerduty/pull/459, traditionally they're not that fast at reviewing PRs, so the timeline is probably on the order of weeks

jewzaam:
1. If it's not expected that the grouping is deleted why would we do that?
2. Did you reach out to PD to see if it is on their radar?  In the past I've raised issues via chat on the website.  I'd rather look at fixing upstream vs hacking around it if we can and understanding how quickly we can get a fix is important for that.  Additionally I don't know the impact of delaying merging this workaround.  This depends on the toil for this (manually disabling the PD service), how frequently it is done, and how long we would be waiting for a fix.

mjlshen:
1. It's definitely a workaround bandaid and based on how we use PagerDuty. Alert Grouping is a manual process that a RL can perform in PagerDuty and leveraged when a cluster is generating a lot of flapping alerts to ease the experience of on-call. When a service is disabled, all outstanding alerts are resolved and then no more alerts are accepted, resulting in Alert Grouping having no effect. When a cluster is removed from limited support, the expectation is that the problem is fixed and - this is just my opinion - but in this case a cluster should be just like all the rest, without Alert Grouping. If there continue to be flapping alerts a RL can go in and enable it again.
2. Ah, I did reach out to them, since it's not a bug with the PagerDuty api and go-pagerduty itself, making a PR is the fastest way to move it forward haha. The crux of the toil is that no one actually has permissions to manually disable the service. A RL would need to go in and manually remove Alert Grouping, which would allow the operator to disable the service. We definitely could make an alert for when this needs to happen? It seemed a bit much when the end result was the same, removing Alert Grouping on the service

PR #203, opened 2022-10-11
feat: Adding probes to csv deployments

Ankit152:
Signed-off-by: Ankit Kurmi <akurmi@redhat.com>

## Description

Added missing probes in the CSV deployments. 
Ref: https://github.com/openshift/pagerduty-operator/blob/master/main.go#L74

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,6 +20,18 @@ spec:
           command:
           - pagerduty-operator
           imagePullPolicy: Always
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: 8081
+            initialDelaySeconds: 15
+            periodSeconds: 20
+          readinessProbe:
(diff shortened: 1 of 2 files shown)


mrbarge:
I gave this a test and it seems to behave fine.
/lgtm

mjlshen:
Probes look good and I'm good to merge this in and take care of this in a follow up:

What is this `manifests/` directory? As far as I can tell boilerplate only acts on the `deploy/` directory. Do `03-service_account.yaml` and `04-role_binding.yaml` need to be moved into `deploy/` and then we can delete `manifests/`?

Not sure if there's additional context @wshearn https://github.com/openshift/pagerduty-operator/commit/28f1374224719529b56e2f6d2ea90456d6866971

mjlshen (approved):
/approve

PR #204, opened 2022-11-15
[OSD-12647] update the operator to be able to set service orchestration

bmeng:
Update the API with the new serviceOrchestration related parameters
Update the main reconcile loop to handle the service orchestration after PD service created

Additional changes needed:
new configmap with pagerduty service orchestration rule created within operator namespace


fahlmant:
@bmeng is there an OSD card for this? And some documentation on Service Orchestration?

bmeng:
hi @fahlmant The card for this PR is in the title OSD-12647
this is the epic: https://issues.redhat.com/browse/SDE-2096 
this is the spike before the implementation: https://issues.redhat.com/browse/OSD-13250 
cc @NautiluX

pkg/pagerduty/service.go
```diff
@@ -362,6 +378,56 @@ func (c *SvcClient) DisableService(data *Data) error {
 	return nil
 }
 
+// ToggleServiceOrchestration enables/disables the service orchestration for a given PD service
+func (c *SvcClient) ToggleServiceOrchestration(data *Data, active bool) error {
```
mjlshen, line 382: This is reasonable short-term, but should this be contributed upstream to https://github.com/PagerDuty/go-pagerduty/tree/main?

controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+
+	if pdData.ServiceID == "" {
+		// PD service has not been created, skip the service orchestration steps
+		return nil
+	}
+	if pdData.ServiceOrchestrationState != "active" {
+		r.reqLogger.Info("enabling the service orchestration")
+		err := pdclient.ToggleServiceOrchestration(pdData, true)
```
mjlshen, line 79: Wouldn't this enable ServiceOrchestration for every single PagerDuty service? (Is that intended? I may have forgotten)

api/v1alpha1/pagerdutyintegration_types.go
```diff
@@ -50,6 +50,13 @@ type PagerDutyIntegrationSpec struct {
 
 	// Name and namespace in the target cluster where the secret is synced.
 	TargetSecretRef corev1.SecretReference `json:"targetSecretRef"`
+
+	ServiceOrchestration ServiceOrchestration `json:"serviceOrchestration"`
```
mjlshen, line 54: Small nit, but can we add description comments to these new structs/fields?

mjlshen (commented):
Looks good overall! Just a couple small comments/questions

controllers/pagerdutyintegration/serviceorchestration_handler.go
bmeng: In pagerdutyintegration_controller.go, this whole `handleServiceOrchestration()` function will be called only when `if pdi.Spec.ServiceOrchestration.Enabled`.
So the goal here is going to enable the service orchestration for all the osd/rosa clusters' PD service which match the selectors defined in the PagerdutyIntegration `osd`.

pkg/pagerduty/service.go
bmeng: I love this idea. :100:  I will put it on my list.

controllers/pagerdutyintegration/serviceorchestration_handler.go
mjlshen: Right, thanks for clarifying - that makes much more sense!

api/v1alpha1/pagerdutyintegration_types.go
bmeng: Added, thanks!

pkg/pagerduty/service.go
bmeng: https://issues.redhat.com/browse/OSD-13926 created here.

mjlshen:
Looks good to me, to pass the validate PR check I think you'll need to rebase and `make container-generate`

mjlshen (approved):
/lgtm

PR #205, opened 2022-11-15
boilerplate update

anispate:
1. $make boilerplate-update
2. Update Go 1.19 and its Dependency bumps as indicated in https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.25.0
3. Found gosec positive therefore, had to update that 
4. $./boilerplate/_lib/container-make ensure-fips
5. $./boilerplate/_lib/container-make generate


mjlshen (approved):
/lgtm
/approve

Thanks for doing this! Make sure to validate this works in stage before the prod promotion

PR #208, opened 2022-11-30
set the enable service orchestration and apply orchestration rule as parallel logic

bmeng:
Set the `ServiceOrchetrationEnabled`(was `ServiceOrchestrationState`) and `ServiceOrchestrationRuleApplied` separately in the cluster configmap to make sure the reconcile loop could handle them separately.

There was a problem before, that it is possible the `ServiceOrchestrationState` set to `active` when the ServiceOrchestration enabled but the ServiceOrchestration rule was not applied due to it is missing at the beginning. And the reconcile loop will not apply it anymore when the rule added since the value set the `active`. It will also be an issue if we want to update the rule in future.

Separated indicator in configmap could call the `ToggleServiceOrchestration` and `ApplyServiceOrchestrationRule` in parallel. 

Also add the related tests.

### DIFF
--- a/controllers/pagerdutyintegration/clusterdeployment_created.go
+++ b/controllers/pagerdutyintegration/clusterdeployment_created.go
@@ -94,7 +94,7 @@ func (r *PagerDutyIntegrationReconciler) handleCreate(pdclient pd.Client, pdi *p
 		r.reqLogger.Info("Creating configmap")
 
 		// save config map
-		newCM := kube.GenerateConfigMap(cd.Namespace, configMapName, pdData.ServiceID, pdData.IntegrationID, pdData.EscalationPolicyID, pdData.ServiceOrchestrationState, false, false)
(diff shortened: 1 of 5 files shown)


mjlshen:
> There was a problem before, that it is possible the ServiceOrchestrationState set to active when the ServiceOrchestration enabled but the ServiceOrchestration rule was not applied due to it is missing at the beginning. And the reconcile loop will not apply it anymore when the rule added since the value set the active. It will also be an issue if we want to update the rule in future.

> Separated indicator in configmap could call the ToggleServiceOrchestration and ApplyServiceOrchestrationRule in parallel.

I think this is a little murky since the ConfigMap isn't really validated. What if a user does `ServiceOrchetrationEnabled: false` and `ServiceOrchestrationRuleApplied: some rule`?

mjlshen (approved):
/lgtm

Thanks for this!

PR #209, opened 2022-12-05
remove the condition check before apply the rule and update the olm template for the CR

bmeng:
Was adding a if condition check on the cluster configmap to reduce the request number to the pagerduty service.

But turns out that it can block the reconcile loop when the orchestration rule got updated. Since the value of `ServiceOrchestrationRuleApplied` will be set to true right after the first time the rule applied.

Remove the check to let the update of orchestration rule configmap could get applied.

Also, update the olm template for the CR to include the service orchestration field.

There will be another PR in app-interface to set the variables.


controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -99,50 +99,48 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
...
+			serviceOrchestrationConfigMap, ServiceOrchestrationDataName)
+		if err != nil {
+			return err
 		}
+
+		pdData.ServiceOrchestrationRules = orchestrationConfig
+
+		r.reqLogger.Info(fmt.Sprintf("apply the service orchestration rules from configmap: %s",
+			orchestrationConfigmapName))
+		err = pdclient.ApplyServiceOrchestrationRule(pdData)
```
mrbarge, line 130: Does this mean it would happen on every reconcile for every cluster? If so, given the `ApplyServiceOrchestrationRule` function seems to itself make two calls (to get the service and to apply the orchestration), I am concerned over the potential that this might lead to potential rate limiting in our organization.

That being said, I don't have better immediate options here other than to halve the number by first retrieving the service orchestration from PD and comparing if it is different to what is defined via the configmap/pdi.

mjlshen:
Should `ServiceOrchestrationRuleApplied` be a status condition then instead of a config map field if it needs to be updated as the ConfigMap changes?

controllers/pagerdutyintegration/serviceorchestration_handler.go
bmeng: The `ApplyServiceOrchestrationRule` will have one api call, which is HTTP PUT.

Based on my testing, it won't be triggered in the reconcile loop every time. Not sure if the `Reconcile()` return is preventing that https://github.com/openshift/pagerduty-operator/blob/master/controllers/pagerdutyintegration/pagerdutyintegration_controller.go#L236 

And also, other func in the controller is not being called everytime as well, like https://github.com/openshift/pagerduty-operator/blob/master/controllers/pagerdutyintegration/clusterdeployment_created.go#L168

bmeng:
As the current test result (and the statement I mentioned above in the comment), remove the condition check will not increasing API calls to the Pagerduty service as how the operator worked. I would like to make minimal change if possible.

@mjlshen WDYT? are we good to merge? 
btw, the app-interface side PR merged which should unblock this as well.

/unhold

mjlshen:
@bmeng Yes, I am ok with this, just wanted to clarify what you were experiencing and what you considered when working on the fix. Thanks!

/lgtm

PR #212, opened 2022-12-13
Update Default OLM Template Value

dustman9000:
This "stringifies" the default bool value for the SERVICE_ORCHESTRATION_ENABLED variable as app-interface expects this to be a string.


fahlmant (approved):
/lgtm
/approve

PR #213, opened 2022-12-17
Add service orchestration configmap template

mrbarge:
# What

- Introduces a new template representing the pagerduty-operator's service orchestration configmap. 

# Why 

This template will be processed by app-interface to deploy the processed template onto Hive shards so that pagerduty-operator can (when enabled) perform service orchestration enablement on managed PD services.

# Questions and answers

_Why is this being introduced as a separate template rather than in the OLM template bundle?_

- I don't believe a processed template for this is a valid resource to go in the OLM catalog bundle. Including it as part of the catalog would require promoting a new release of the pagerduty operator every time we want to roll out service orchestration rule updates. Rule updates, especially if we grow them in the future, should stand independent of the operator.

_Why can this not be deployed in managed-cluster-config?_

- It could, but we lose the ability to then have unique configurations for integration, staging and production, because all Hive shards are production clusters. We can and should test out rule updates in staging before deploying them to Production.
- app-interface should always be a preferred deployment mechanism for Hive-related configurations.

# How will this work?

An associated app-interface change to pagerduty-operator's SaaS file will process the template and populate the parameters with expected values (which will be similarly managed in app-interface and can be made environment-specific in future if needed):

https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/54352

# Validation

This has been tested locally:

```
oc process --local -f 00-osd-serviceorchestration.configmap.yaml.tmpl -o yaml SERVICE_ORCHESTRATION_RULE_CONFIGMAP=osd-serviceorchestration SERVICE_ORCHESTRATION_RULES="$ruleData"
```

# Refs

[OSD-12648](https://issues.redhat.com//browse/OSD-12648)

### DIFF
--- a/hack/templates/00-osd-serviceorchestration.configmap.yaml.tmpl
+++ b/hack/templates/00-osd-serviceorchestration.configmap.yaml.tmpl
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: osd-serviceorchestration-template
+parameters:
+- name: SERVICE_ORCHESTRATION_RULES
+  required: true
+- name: SERVICE_ORCHESTRATION_RULE_CONFIGMAP
+  required: true
+objects:
+- apiVersion: v1
+  data:
+    service-orchestration.json: ${SERVICE_ORCHESTRATION_RULES}
+  kind: ConfigMap
+  metadata:
+    name: ${SERVICE_ORCHESTRATION_RULE_CONFIGMAP}
+    namespace: pagerduty-operator
+    labels:
+      api.openshift.com/managed: "true"


mjlshen (commented):
I do like this approach and this is similar to what was done in https://github.com/openshift/aws-vpce-operator/pull/66

However, is there some limitation here that requires the use of a separate template instead of just having the ConfigMap itself in the single OLM template?

mrbarge:
> However, is there some limitation here that requires the use of a separate template instead of just having the ConfigMap itself in the single OLM template?

It's not so much a limitation, but rather I'm thinking that if it goes in the OLM template then it becomes part of the catalog image, which then means that the configuration is wedded to the version of the PD operator itself and therefore also the promotion process of the operator.

Treating it separately means that the configuration can evolve separately to the operator and not require the operator's latest code base to be pushed out to prod alongside future rule changes. (A "Promotion" process of the rule configuration to prod can still occur, but would be managed entirely through app interface MRs)

mjlshen:
> Treating it separately means that the configuration can evolve separately to the operator and not require the operator's latest code base to be pushed out to prod alongside future rule changes. (A "Promotion" process of the rule configuration to prod can still occur, but would be managed entirely through app interface MRs)

Yeah, agreed, I thought it would have been no different if the ConfigMap with templated values was in the big OLM template, but this would make it even more explicit in app-interface than the alternative

/lgtm

PR #214, opened 2022-12-19
write service orchestration rule data to clusterconfigmap directly

bmeng:
write the contents of the service orchestration rule to the clusterconfigmap

so that, we can skip the `ApplyServiceOrchestrationRule()` call if it has been applied already.

The logic will be:

- Read the `serviceorchestrationconfigmap` for pagerduty-operator to configure the service orchestration
- Read the `clusterconfigmap` which includes the applied orchestration rule
- Apply if value in `clusterconfigmap` does not match the `serviceorchestrationconfigmap`
- Update the `clusterconfigmap` after apply

### DIFF
--- a/controllers/pagerdutyintegration/serviceorchestration_handler.go
+++ b/controllers/pagerdutyintegration/serviceorchestration_handler.go
@@ -11,8 +11,6 @@ import (
 	"github.com/openshift/pagerduty-operator/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"reflect"
 )
@@ -71,18 +69,12 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
 		return err
 	}
 
-	pdiSelector, err := metav1.LabelSelectorAsSelector(&pdi.Spec.ClusterDeploymentSelector)
-	if err != nil {
-		return err
-	}
-
 	if pdData.ServiceID == "" {
 		// PD service has not been created, skip the service orchestration steps
 		return nil
 	}
 
 	if !pdData.ServiceOrchestrationEnabled {
-
 		r.reqLogger.Info("enabling the service orchestration")
 		err = pdclient.ToggleServiceOrchestration(pdData, true)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -100,13 +99,13 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
...
-		Name:      pdi.Spec.ServiceOrchestration.RuleConfigConfigMapRef.Name,
+		Name:      orchestrationConfigmapName,
 		Namespace: pdi.Spec.ServiceOrchestration.RuleConfigConfigMapRef.Namespace,
 	}
 
-	ruleConfigmap := &corev1.ConfigMap{}
+	orchestrationRuleConfigmap := &corev1.ConfigMap{}
 
-	err = r.Get(context.TODO(), serviceOrchestrationConfigMap, ruleConfigmap)
+	err = r.Get(context.TODO(), serviceOrchestrationConfigMap, orchestrationRuleConfigmap)
```
mrbarge, line 108: This feels redundant as the `utils.LoadConfigMapData` will be loading the configmap anyway, and reporting it if fails. It seems only needed for the sake of `GetLabels()` check, which as I've noted in a separate comment I'm not sure is all that essential.

controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -116,29 +115,33 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
...
-			r.reqLogger.Error(err, "Error updating PagerDuty cluster config", "Name",
-				clusterConfigmapName)
-			return err
+		if pdData.ServiceOrchestrationRuleApplied != orchestrationRuleConfigData {
+			// Apply rule for new service
+			pdData.ServiceOrchestrationRuleApplied = orchestrationRuleConfigData
+
+			r.reqLogger.Info(fmt.Sprintf("applying the service orchestration rules from configmap: %s",
+				orchestrationConfigmapName))
+			err = pdclient.ApplyServiceOrchestrationRule(pdData)
```
mrbarge, line 132: Just wondering, as I've not tested the code - does this result in the existing rule(s) being entirely replaced with the new rule? Just want to make sure this doesn't result in a second new rule being added to the service's ruleset whilst leaving the first still there.

I ask because the PD API [0] is unclear to me when I read it as to whether the orchestration rule id is required in order to update-in-place an existing rule. And the id would not be known, nor passed through to the PUT, because it's using the definition in the orchestration rule configmap for the body of the PUT.

[0] https://developer.pagerduty.com/api-reference/aeaec09888036-update-the-service-orchestration-for-a-service

controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -116,29 +115,33 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
 		return err
 	}
 
-	if pdiSelector.Matches(labels.Set(ruleConfigmap.GetLabels())) {
-		orchestrationConfig, err := utils.LoadConfigMapData(r.Client,
+	var orchestrationRuleConfigData string
+	if pdiSelector.Matches(labels.Set(orchestrationRuleConfigmap.GetLabels())) {
```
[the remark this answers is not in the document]
bmeng: The label selector checking is not added here, but in https://github.com/openshift/pagerduty-operator/blob/master/controllers/pagerdutyintegration/event_handlers.go#L231

The controller will watch for the configmaps which have the same label with the PDI.

It seems make sense to me to limit the configmap which match the labels to trigger the reconcile.

I will test if remove the if condition here, it will work the same.

controllers/pagerdutyintegration/serviceorchestration_handler.go
bmeng: Yes, it will update the existing rule (not sure if we should call it existing rule, but there will be only one service orchestration rule for each service.)

mrbarge: Excellent, thanks for checking!

controllers/pagerdutyintegration/serviceorchestration_handler.go
bmeng: It is still needed? I think the `utils.LoadConfigMapData` func cannot handle the configmap absent situation.

controllers/pagerdutyintegration/serviceorchestration_handler.go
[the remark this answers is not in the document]
bmeng: Tested with the `if pdiSelector.Matches(labels.Set(orchestrationRuleConfigmap.GetLabels()))` exists in `serviceorchestration_handler.go` or not.

When the if condition exists in `serviceorchestration_handler.go`
- When the configmap has the matched label, 
It will call the func `ApplyServiceOrchestration()` on new clusterdeployment creation and on service orchestration configmap updated.
- When the configmap does not have the matched label,
It will **not** call the func `ApplyServiceOrchestration()` on new clusterdeployment creation and on service orchestration configmap updated

When the if condition does not exist in `serviceorchestration_handler.go`
- When the configmap has the matched label, 
It will call the func `ApplyServiceOrchestration()` on new clusterdeployment creation and on service orchestration configmap updated.
- When the configmap does not have the matched label,
It will call the func `ApplyServiceOrchestration()` on new clusterdeployment creation but not on service orchestration configmap updated

Consider the label in the configmap has already defined/applied. 
The if logic in `serviceorchestration_handler.go` can be removed.

controllers/pagerdutyintegration/serviceorchestration_handler.go
mrbarge: The first thing `utils.LoadConfigMapData` does is do a `Get` on the configmap and return an error if that fails, so it should handle it fine.

mrbarge:
Though not immediately related, there's some additional concerns I've observed whilst testing this that would be good to resolve in this PR.

The operator is cluster-scoped, this means that doing the `Watch` on `ConfigMap` is going to try and enqueue requests for every possible ConfigMap event on the cluster - which is a lot. And the first thing it does is in the event handler is `List` and churn through all the PDIs on the cluster and determine if the configmap matches labels. This makes the operator extremely resource heavy for 99% of things it won't ever enqueue a request for.

I would propose that, given given that the only `ConfigMap` the operator ever needs to care about is the one in the operator's namespace, that the very first thing the `toRequests` should do is check if the namespace is what we expect. If it is not, do not enqueue any requests.

controllers/pagerdutyintegration/serviceorchestration_handler.go
bmeng: Tested and confirmed it works as before after remove this. Thanks!

bmeng:
Thanks for the review, added the configmap namespace checking before enqueue the request. And tested working well.

mrbarge:
Thanks @bmeng , I have given it a test as well and can confirm it is behaving well, so lgtm from my side.

mrbarge:
Now that I'm a reviewer I'll flex that muscle slightly :)
/lgtm

mjlshen (approved):
/approve

PR #219, opened 2023-02-19
Add more context to errors with pagerduty client interaction

mrbarge:
# What it does
- Wraps errors stemming from pagerduty-client interactions with more context regarding what action was caused to trigger the error.
- Log timestamps in human-readable form rather than epoch seconds (took inspiration from [this](https://github.com/openshift/cloud-ingress-operator/pull/280) PR)

# Why
The errors returned from the pagerduty client library rarely give any context as to what API interaction took place. This makes it confusing or misleading when reviewing PDO logs to understand why an error is occurring. This PR wraps each error with additional context surrounding the call so that when it is passed up the chain there is a clearer picture of why the failure is occurring.

# Refs

[OSD-14350](https://issues.redhat.com//browse/OSD-14350)

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/openshift/pagerduty-operator/controllers/pagerdutyintegration"
 	"github.com/openshift/pagerduty-operator/pkg/localmetrics"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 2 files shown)


mjlshen (approved):
/hold
/lgtm

Thanks so much for doing this! Just holding since it's the weekend, feel free to unhold whenever

PR #224, opened 2023-04-28
OSD-16066 alert grouping

yithian:
When defined as part of a pagerdutyintegration spec, the `Spec.AlertGroupingParameters.Type` and `Spec.AlertGroupingParameters.Config.Timeout` values will enable alert grouping for clusters' Pager Duty services.

New clusters' Pager Duty services should have alert grouping enabled by the change to the `CreateService()` function. Existing clusters should be updated via the `pagerdutyintegration` `Reconcile()` loop: the `Spec.AlertGroupingParameters.Type` and `Spec.AlertGroupingParameters.Config.Timeout` values will not match what's in the cluster's Pager Duty service configmap and will perform an update.


pkg/pagerduty/service.go
```diff
@@ -202,6 +214,19 @@ func (data *Data) ParseClusterConfig(osc client.Client, namespace string, cmName
...
+	agto, err := getConfigMapKey(pdAPIConfigMap.Data, "ALERT_GROUPING_TIMEOUT")
+	if err != nil {
+		// if the alert grouping timeout isn't set in the configmap, default to 0
+		agto = "0"
+	}
+	agtou64, err := strconv.ParseUint(agto, 10, 64)
+	if err != nil {
+		return err
+	}
+	data.AlertGroupingTimeout = uint(agtou64)
```
mjlshen, line 228: Does this defaulting of `ALERT_GROUPING_TYPE`, and `ALERT_GROUPING_TIMEOUT` overwrite what gets set from the `PagerDutyIntegration` in all cases (I think so)?

mjlshen (commented):
Looks good overall, just one clarifying question on my end

pkg/pagerduty/service.go
yithian: The values in the configmap shouldn't be directly set from the values in the pagerdutyintegration resource anywhere. The indented flow for existing clusters is:

1. alert grouping gets configured in the pagerdutyintegration
2. the current state of cluster X's Pager Duty service gets loaded up from its configmap
2a. if configmap doesn't have entries for alert grouping, then initialize them to something sane to avoid breakage in 3
3. compare the desired state (from the pagerdutyintegration) to the current state (from the configmap data)
4. if they're different, update the service to the values from the pagerdutyintegration
5. update the values in the configmap to those from the new current state of the Pager Duty service

My brain was stretching a bit to hold all the changes in my head at the same time, so if this won't work or isn't what this PR will do, let me know!

mjlshen: No that makes sense, I forgot the current state was stored in the Configmap, not the expected state is all - thanks for the explanation!

PR #229, opened 2023-06-21
OSD-16066 Ensure alert grouping gets applied to existing clusters

yithian:
This updates the `handleUpdate()` function (and section of the reconcile loop where it's called) to ensure that existing clusters have alert grouping enabled when configured via a pagerdutyintegration. There are unit tests for the reconcile loop that should confirm this behavior

### DIFF
--- a/controllers/pagerdutyintegration/handleupdate.go
+++ b/controllers/pagerdutyintegration/handleupdate.go
@@ -59,7 +59,10 @@ func (r *PagerDutyIntegrationReconciler) handleUpdate(pdclient pd.Client, pdi *p
 
 		cm.Data["ALERT_GROUPING_TYPE"] = pdi.Spec.AlertGroupingParameters.Type
 		cm.Data["ALERT_GROUPING_TIMEOUT"] = fmt.Sprintf("%d", pdi.Spec.AlertGroupingParameters.Config.Timeout)
-		r.Client.Update(context.TODO(), cm)
+		err = r.Client.Update(context.TODO(), cm)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/controllers/pagerdutyintegration/handleupdate.go
+++ b/controllers/pagerdutyintegration/handleupdate.go
@@ -15,8 +15,13 @@
 package pagerdutyintegration
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/openshift/pagerduty-operator/config"
 	pd "github.com/openshift/pagerduty-operator/pkg/pagerduty"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	pagerdutyv1alpha1 "github.com/openshift/pagerduty-operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


controllers/pagerdutyintegration/handleupdate.go
```diff
@@ -28,20 +33,33 @@ func (r *PagerDutyIntegrationReconciler) handleUpdate(pdclient pd.Client, pdi *p
...
-	pdData, err := pd.NewData(pdi, cd.Spec.ClusterMetadata.ClusterID, cd.Spec.BaseDomain)
-	if err != nil {
-		return err
-	}
-	err = pdData.ParseClusterConfig(r.Client, cd.ObjectMeta.Namespace, configMapName)
+	cm := &corev1.ConfigMap{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: cd.ObjectMeta.Namespace, Name: configMapName}, cm)
 	if err != nil {
-		return err
+		return nil // requeue and wait for the configmap to be created
```
mjlshen, line 39: I believe returning an error is what would cause this to requeue right? https://github.com/openshift/pagerduty-operator/pull/229/files#diff-cecb7e9ffd55b2eb077a62f26087f4dc1a582b736b8cefc9771ef67840350efbR238-R240

controllers/pagerdutyintegration/handleupdate.go
```diff
@@ -28,20 +33,33 @@ func (r *PagerDutyIntegrationReconciler) handleUpdate(pdclient pd.Client, pdi *p
...
+		}
+
 		err = pdclient.UpdateAlertGrouping(pdData)
 		if err != nil {
 			return err
 		}
+
+		cm.Data["ALERT_GROUPING_TYPE"] = pdi.Spec.AlertGroupingParameters.Type
+		cm.Data["ALERT_GROUPING_TIMEOUT"] = fmt.Sprintf("%d", pdi.Spec.AlertGroupingParameters.Config.Timeout)
+		r.Client.Update(context.TODO(), cm)
```
mjlshen, line 62: We should handle the error for this one too to make sure it gets requeued if this fails

mjlshen (changes requested):

controllers/pagerdutyintegration/handleupdate.go
yithian: If it returns the error, it causes the entire reconcile loop to return the error, which causes the following test case to fail:

```
--- FAIL: TestReconcilePagerDutyIntegration (0.05s)
    --- FAIL: TestReconcilePagerDutyIntegration/Test_Managed,_No_Finalizer,_Not_Deleting,_PD_Not_Setup (0.00s)
        pagerdutyintegration_controller_test.go:1120:
                Error Trace:    pagerdutyintegration_controller_test.go:1120
                Error:          Expected nil, but got: pagerdutyintegration.pdiReconcileErrors{(*errors.StatusError)(0xc0001a0be0)}
                Test:           TestReconcilePagerDutyIntegration/Test_Managed,_No_Finalizer,_Not_Deleting,_PD_Not_Setup
                Messages:       Unexpected Error with Reconcile (1 of 3)
```

I erred ;) on the side of not changing existing behavior; do we want to do something different for this test case?

mjlshen: I guess what I'm worried about is that the reconcile loop won't get requeued if we don't return an error here. Maybe it's not a big deal since the reconcile loop will run any time a clusterdeployment gets added so we shouldn't get into a state where something's not configured and a reconcile loop won't run for "a while" (I'm not sure)?

yithian: I'm fuzzy on this, to be honest. From what I can understand of [the test code](https://github.com/openshift/pagerduty-operator/blob/master/controllers/pagerdutyintegration/pagerdutyintegration_controller_test.go#L1046-L1068), it looks like pagerduty-operator expects to reconcile multiple times to "finalize" things?

I can change that test case to expect an error if you think that's better, though!

mjlshen: I think that's ok then, it is true that the `Reconcile()` function won't get automatically called unless this returns an error, but seems like it's safe to assume that it'll get called a bunch anyway. Sorry for the delay here!

mjlshen (approved):
/lgtm

PR #232, opened 2023-07-03
Fix issue where AlertGroupingParameters was nil

geowa4:
When nil, it would seg fault.

[OHSS-24073](https://issues.redhat.com//browse/OHSS-24073)


mjlshen (approved):
/lgtm

PR #235, opened 2023-08-08
feat: allow a separate event orchestration for RH infra clusters

typeid:
This PR is a requirement for https://issues.redhat.com/browse/OSD-17861 - after merging this, we will need to update `app-interface` with the according `rh-infra-service-orchestration.json`

**What?**

This PR enables us to pass a second event orchestration configuration that will be applied to Red Hat infrastructure clusters (currently targets hypershift SC and MC). 

**Why?**

We want to be able to react to infrastructure cluster alerts before other alerts, so that we can fix issues that possibly impact a large number of other clusters quickly. For this, we need to be able to customize infrastructure cluster alerts, which can be done using the following event orchestration configuration applied to only infrastructure clusters:

```
  rh-infra-service-orchestration.json: |
    {
    "orchestration_path": {
    "catch_all": {
        "actions": {}
    },
    "sets": [
      {
        "id": "start",
        "rules": [
          {
            "actions": {
              "suppress": true
            },
            "conditions": [
              {
                "expression": "event.severity matches 'Warning'"
              }
            ]
          },
          {
            "actions": {
              "priority": "PQF279U"
            },
            "conditions": []
          }
        ]
      }
    ],
        "type": "service"
    }
    }
```

Resulting alerts will have a priority tag:
image

**Tested using CRC and a trial PD account:**
- PD Operator behaves as expected for standard clusters
- PD Operator applies `rh-infra-service-orchestration.json` for MCs and SCs.

Note: I will open up a follow-up PR to update the README for CRC and fix some outdated guidance.

### DIFF
--- a/controllers/pagerdutyintegration/serviceorchestration_handler.go
+++ b/controllers/pagerdutyintegration/serviceorchestration_handler.go
@@ -3,6 +3,8 @@ package pagerdutyintegration
 import (
 	"context"
 	"fmt"
+	"reflect"
+
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	pagerdutyv1alpha1 "github.com/openshift/pagerduty-operator/api/v1alpha1"
 	"github.com/openshift/pagerduty-operator/config"
@@ -12,11 +14,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	"reflect"
 )
 
 const (
-	ServiceOrchestrationDataName = "service-orchestration.json"
+	StandardServiceOrchestrationDataName    = "service-orchestration.json"
+	RedHatInfraServiceOrchestrationDataName = "rh-infra-service-orchestration.json"
 )
 
 // handleServiceOrchestration enables and applies the service orchestration rule to the PD service if it is enabled in PDI
@@ -96,14 +98,31 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
 		Namespace: pdi.Spec.ServiceOrchestration.RuleConfigConfigMapRef.Namespace,
 	}
 
-	orchestrationRuleConfigData, err := utils.LoadConfigMapData(r.Client,
-		serviceOrchestrationConfigMap, ServiceOrchestrationDataName)
-	if errors.IsNotFound(err) {
-		r.reqLogger.Info("service orchestration configmap rule not found, skip the following steps")
-		localmetrics.UpdateMetricPagerDutyServiceOrchestrationFailure(1, pdi.Name)
-		return nil
-	} else if err != nil {
-		return err
+	orchestrationRuleConfigData := ""
+
+	if utils.IsRedHatHyperShiftInfrastructure(cd) {
+		orchestrationRuleConfigData, err = utils.LoadConfigMapData(r.Client,
+			serviceOrchestrationConfigMap, RedHatInfraServiceOrchestrationDataName)
+
+		if errors.IsNotFound(err) {
+			r.reqLogger.Info(fmt.Sprintf("found no service orchestration configmap rule for '%s', skipping next steps", RedHatInfraServiceOrchestrationDataName))
+			localmetrics.UpdateMetricPagerDutyServiceOrchestrationFailure(1, pdi.Name)
+			return nil
+		} else if err != nil {
+			return err
+		}
+	} else {
+		orchestrationRuleConfigData, err = utils.LoadConfigMapData(r.Client,
+			serviceOrchestrationConfigMap, StandardServiceOrchestrationDataName)
+
+		if errors.IsNotFound(err) {
+			r.reqLogger.Info(fmt.Sprintf("found no service orchestration configmap rule for '%s', skipping next steps", StandardServiceOrchestrationDataName))
+			localmetrics.UpdateMetricPagerDutyServiceOrchestrationFailure(1, pdi.Name)
+			return nil
+		} else if err != nil {
+			return err
+		}
+
 	}
 
 	if pdData.ServiceOrchestrationRuleApplied != orchestrationRuleConfigData {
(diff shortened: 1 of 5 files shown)


hack/templates/00-osd-serviceorchestration.configmap.yaml.tmpl
```diff
@@ -11,6 +11,7 @@ objects:
 - apiVersion: v1
   data:
     service-orchestration.json: ${SERVICE_ORCHESTRATION_RULES}
+    rh-infra-service-orchestration.json: ${RH_INFRA_SERVICE_ORCHESTRATION_RULES}
```
jewzaam, line 14: Missing definition of `RH_INFRA_SERVICE_ORCHESTRATION_RULES` in list of parameters in this template.

pkg/utils/utils.go
```diff
@@ -141,3 +141,31 @@ func GetClusterID(cd *hivev1.ClusterDeployment) string {
 		return cd.Spec.ClusterName
 	}
 }
+
+// IsRedHatHyperShiftInfrastructure returns whether or not a cluster is part of the Red Hat HyperShift infrastructure
+func IsRedHatHyperShiftInfrastructure(cd *hivev1.ClusterDeployment) bool {
```
jewzaam, line 146: Function name isn't consistent with configuration.  Recommend:

```suggestion
func IsRedHatInfrastructure(cd *hivev1.ClusterDeployment) bool {
```

pkg/utils/utils.go
```diff
@@ -141,3 +141,31 @@ func GetClusterID(cd *hivev1.ClusterDeployment) string {
...
+	// hypershiftClusterTypeLabel is the annotation key for the hypershift cluster type
+	hypershiftClusterTypeLabel := "ext-hypershift.openshift.io/cluster-type"
+
+	// hypershiftInfrastructureClusterTypes are the values of the hypershiftClusterTypeLabel key
+	// that represent critical Red Hat infrastructure clusters
+	hypershiftInfrastructureClusterTypes := []string{"service-cluster", "management-cluster"}
```
jewzaam, line 152: This is binding PD operator to labels set by OSDFM, which I think we have to do.  Risk of upstream changes that could result in this no longer being set.  Suggest talking w/ OSDFM team to identify if there is a better solution.  Perhaps a PD specific label, even if set by OSDFM, that is clearly not something that can be changed without some impact downstream.

Suggest:
* key = ext-pagerduty.openshift.io/rh-infra
* value = 'true'

jewzaam (changes requested):
Include MR for the rh-infra service orchestration rules in JIRA for reference.  I assume that's where the value of `RH_INFRA_SERVICE_ORCHESTRATION_RULES` will be set.

hack/templates/00-osd-serviceorchestration.configmap.yaml.tmpl
typeid: Good catch, thanks. Done.

pkg/utils/utils.go
typeid: Your suggestion to have an  `ext-pagerduty.openshift.io/rh-infra` annotation has been implemented in OSDFM as part of https://issues.redhat.com/browse/OCM-3176. I updated this PR accordingly, but am putting it on hold and will release once the change is fully propagated on staging.

typeid:
/hold

Waiting for https://issues.redhat.com/browse/OCM-3176 change to propagate on staging.

typeid:
/unhold

- `ext-pagerduty.openshift.io/rh-infra` has been propagated on staging. 
- `rh-infra-service-orchestration` is set in app-interface

Ready from my side to merge this to staging! :)

controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -96,14 +98,31 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
...
-		r.reqLogger.Info("service orchestration configmap rule not found, skip the following steps")
-		localmetrics.UpdateMetricPagerDutyServiceOrchestrationFailure(1, pdi.Name)
-		return nil
-	} else if err != nil {
-		return err
+	orchestrationRuleConfigData := ""
+
+	if utils.IsRedHatInfrastructure(cd) {
+		orchestrationRuleConfigData, err = utils.LoadConfigMapData(r.Client,
+			serviceOrchestrationConfigMap, RedHatInfraServiceOrchestrationDataName)
```
mjlshen, line 105: I'd prefer this to be on one line, don't think it's easier to read this way

mjlshen (approved):
/approve

A couple nits to consider

controllers/pagerdutyintegration/serviceorchestration_handler.go
```diff
@@ -96,14 +98,31 @@ func (r *PagerDutyIntegrationReconciler) handleServiceOrchestration(pdclient pd.
...
 
-	orchestrationRuleConfigData, err := utils.LoadConfigMapData(r.Client,
-		serviceOrchestrationConfigMap, ServiceOrchestrationDataName)
-	if errors.IsNotFound(err) {
-		r.reqLogger.Info("service orchestration configmap rule not found, skip the following steps")
-		localmetrics.UpdateMetricPagerDutyServiceOrchestrationFailure(1, pdi.Name)
-		return nil
-	} else if err != nil {
-		return err
+	orchestrationRuleConfigData := ""
```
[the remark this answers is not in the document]
typeid: `orchestrationRuleConfigData` is set in the if/else and then re-applied if it isn't already applied later L128/130.

controllers/pagerdutyintegration/serviceorchestration_handler.go
typeid: I kept the formatting as it was before, it just moved into the if/else making it even longer. 
Personally I think the line split version is easier to read, but that's just personal preference.

mjlshen (commented):
/lgtm

controllers/pagerdutyintegration/serviceorchestration_handler.go
mjlshen: Sure, no problem, it's a matter of personal preference haha

PR #241, opened 2023-09-08
OSD-17911 alert grouping bugfix

yithian:
When parsing a cluster config for a PagerDuty Data, skip past the alert
grouping parameters. This will cause the Data to keep the alert grouping
parameters from the pagerdutyintegration (set in NewData()), which is
used for the update query to Pager Duty.

The configmap values for alert grouping will be updated after the update
query is successful, and are thus only used to check if an update needs
to be made.


mjlshen (approved):
/lgtm

PR #243, opened 2023-09-25
fix(OSD-15493): service orchestrations not setup correctly for some PD services

typeid:
Sometimes, PDO does not correctly apply the service orchestration (filtering out warning alerts) to a service. 

**The issue:** PDO does not query the service orchestration data from PD, instead it caches the "applied" data for every service in a configmap. This cache is inconsistent due to a bug.

**The bug:** the cached data was inconsistent with the actual applied pagerduty data, as we don't properly check if a HTTP request succeeds (documentation of `http.DefaultClient.Do)` :
```
An error is returned if caused by client policy (such as CheckRedirect), or failure to speak HTTP (such as a network connectivity problem). A non-2xx status code doesn't cause an error.
```

We only check if an error occurs, but don't check the returned status code of the http request.

**The fix:** we only update the configmap if the HTTP request properly succeeds (2xx code).

https://issues.redhat.com/browse/OSD-15493

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -8,7 +8,7 @@ COPY . .
 RUN make go-build
 
 ####
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8-1037
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8-1072
 
 ENV USER_UID=1001 \
     USER_NAME=pagerduty-operator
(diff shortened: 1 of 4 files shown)


mjlshen (approved):
/approve

jewzaam:
Some discussion about metric for the failure.  But the side effect of no metric or alert is on-call is paged when this isn't configured in PD.  Not a great experience but it means it isn't lost.  We'll track some new work around PDO observability but it will not block this PR.

/lgtm

PR #244, opened 2023-10-04
Add variable resolve message for limited support

bdmiller3:
Adds an additional summary message for resolving alerts due to limited support

for [OSD-18059](https://issues.redhat.com//browse/OSD-18059)

### DIFF
--- a/pkg/pagerduty/service.go
+++ b/pkg/pagerduty/service.go
@@ -529,7 +529,7 @@ func (c *SvcClient) resolvePendingIncidents(data *Data) error {
 					alert.Integration.ID, incident.ID, data.ServiceID, err)
 			}
 
-			err = c.resolveAlert(integration.IntegrationKey, alert.AlertKey)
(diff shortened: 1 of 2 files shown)


pkg/pagerduty/service.go
mjlshen, line 1: I think this is clearer if we pass a `summary string` into `resolvePendingIncidents` and `resolveAlert`.

Then, we can modify the locations `resolvePendingIncidents` is called during
* [Delete](https://github.com/openshift/pagerduty-operator/blob/66a2e4a135c7b4b427b2220b6958b2847361035e/pkg/pagerduty/service.go#L350)
* [Disable (Limited Support)](https://github.com/openshift/pagerduty-operator/blob/66a2e4a135c7b4b427b2220b6958b2847361035e/pkg/pagerduty/service.go#L392)

workflows accordingly. I think the way this is now could introduce a different source of confusion if a cluster in limited support is getting deleted.

mjlshen (changes requested):

mjlshen (approved):
/lgtm

PR #250, opened 2023-11-14
OSD-18468 - template for fedramp removal

theautoroboto:
We are in the process of removing PagerDuty from FedRAMP.  This is the first of a few PRs that will occur over the next few months.  This phase will remove PD from all non-production clusters while production clusters will remain active with pager duty.  We will target this new template for fedramp.  Once we have fully transitioned from PD/DMS, this template will be removed.

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.fedramp.yaml
+++ b/hack/olm-registry/olm-artifacts-template.fedramp.yaml
@@ -167,6 +167,9 @@ objects:
       - key: ext-managed.openshift.io/noalerts
         operator: NotIn
         values: ["true"]
+      - key: api.openshift.com/fedramp
+        operator: In
+        values: ["true"]
       - key: api.openshift.com/environment
         operator: NotIn
         values:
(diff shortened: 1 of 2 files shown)


mjlshen (approved):
/lgtm

hack/olm-registry/olm-artifacts-template.fedramp.yaml
```diff
@@ -0,0 +1,239 @@
...
+        operator: NotIn
+        values: ["nightly"]
+      # ignore CD w/ "legacy" noalerts label
+      - key: api.openshift.com/noalerts
+        operator: NotIn
+        values: ["true"]
+      # ignore CD w/ ext noalerts label
+      - key: ext-managed.openshift.io/noalerts
+        operator: NotIn
+        values: ["true"]
```
mjlshen, line 112: Could we add a selector so that this OLM template only selects FedRAMP clusters?

And likewise for the existing OLM template to exclude FedRAMP clusters?

theautoroboto: Updated, thanks

mjlshen (approved):
/unhold
/lgtm

PR #259, opened 2024-01-25
OSD-20576 - reusing the existing integration if there is one instead of creating a new one

Nikokolas3270:


pkg/pagerduty/service.go
```diff
@@ -319,12 +319,18 @@ func (c *SvcClient) CreateService(data *Data) (string, error) {
 		if newSvc == nil {
 			return "", fmt.Errorf("creating service %v returned an empty response: %w", clusterService.Name, err)
 		}
+
+		err = nil
```
typeid, line 323: **Question:** Is there any value in resetting the error here?

pkg/pagerduty/service.go
```diff
@@ -319,12 +319,18 @@ func (c *SvcClient) CreateService(data *Data) (string, error) {
...
-	data.IntegrationID, err = c.createIntegration(newSvc.ID, "V4 Alertmanager", "events_api_v2_inbound_integration")
-	if err != nil {
-		return "", fmt.Errorf("unable to create integration for service %v: %w", newSvc.ID, err)
+	if len(newSvc.Integrations) == 0 {
+		data.IntegrationID, err = c.createIntegration(newSvc.ID, "V4 Alertmanager", "events_api_v2_inbound_integration")
+		if err != nil {
+			return "", fmt.Errorf("unable to create integration for service %v: %w", newSvc.ID, err)
+		}
+	} else {
+		data.IntegrationID = newSvc.Integrations[0].ID
```
typeid, line 333: **Suggestion:** We should probably differentiate here between `len(newSvc.Integrations) == 1` and `len(newSvc.Integrations) > 1`. If there's more than 1, we probably want to log a warning.
If a service has multiple integrations, are we sure the first one is the correct one?

typeid (commented):
What was the actual root cause for the loop-create infinite integrations? 

It looks like we needed to be in a situation where the `cd` had no `DeletionTimestamp` (see [here](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/pagerdutyintegration_controller.go#L211))

Then we end up [here](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/clusterdeployment_created.go#L82-L84)

So there seems to have been two options that could have caused continuous creations of integrations:
a) `ParseClusterConfig` was constantly erroring
b) the `pdData.ServiceID` was never updated: this would be caused by erroring somewhere [during the update/creation of the configmap](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/clusterdeployment_created.go#L97-L112)

pkg/pagerduty/service.go
Nikokolas3270: We could check againt the integration type... But it is a string... and the string returned by the PagerDuty API call is NOT the one provided when creating the Integration object

pkg/pagerduty/service.go
Nikokolas3270: Yes, there is. `err` is eventually returned later. But yeah this is not elegant; will figure something out.

pkg/pagerduty/service.go
```diff
@@ -322,14 +326,34 @@ func (c *SvcClient) CreateService(data *Data) (string, error) {
...
-	if err != nil {
-		return "", fmt.Errorf("unable to create integration for service %v: %w", newSvc.ID, err)
+	var existingIntegration *pdApi.Integration
+
+	for k := range newSvc.Integrations {
+		integration := &newSvc.Integrations[k]
+
+		// The _reference suffix is added to the expected type as we are not manipulating a whole integration object there.
+		// See: https://developer.pagerduty.com/api-reference/165ad96a22ffd-get-a-service
+		if integration.Type == integrationType+"_reference" {
```
typeid, line 336: Can we make `integrationType+"_reference"` a new variable to make it more descriptive? We'd also not need to re-construct it in a loop, and it makes the core loop logic easier to read. 
We can also use the ID instead of the pointer.

What do you think about something like this:

```
	// The _reference suffix is added to the expected type as we are not manipulating a whole integration object there.
	// See: https://developer.pagerduty.com/api-reference/165ad96a22ffd-get-a-service
	expectedIntegrationType := integrationType + "_reference"

	var existingIntegrationID string
	for _, integration := range newSvc.Integrations {
		if integration.Type == expectedIntegrationType {
			if existingIntegrationID == "" {
				existingIntegrationID = integration.ID
			} else {
				log.Log.WithValues("data.ServiceID", data.ServiceID, "integrationType", integrationType, "integration.ID", integration.ID).Info("service already has multiple integrations with the expected type, using first result")
			}
		}
	}

	if existingIntegrationID == "" {
		data.IntegrationID, err = c.createIntegration(newSvc.ID, integrationName, integrationType)
		if err != nil {
			return "", fmt.Errorf("unable to create integration for service %v: %w", newSvc.ID, err)
		}
	} else {
		data.IntegrationID = existingIntegrationID
	}
```

pkg/pagerduty/service.go
```diff
@@ -303,7 +307,7 @@ func (c *SvcClient) CreateService(data *Data) (string, error) {
 		lso.Query = clusterService.Name
 		currentSvcs, newerr := c.PdClient.ListServices(lso)
 		if newerr != nil {
-			return "", fmt.Errorf("unable to list services with name %v: %w", clusterService.Name, err)
+			return "", fmt.Errorf("unable to list services with name %v: %w", clusterService.Name, newerr)
```
typeid, line 310: good catch

Nikokolas3270 (commented):
Thanks for the review; taking comments into account.

Nikokolas3270:
> What was the actual root cause for the loop-create infinite integrations?
> 
> It looks like we needed to be in a situation where the `cd` had no `DeletionTimestamp` (see [here](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/pagerdutyintegration_controller.go#L211))
> 
> Then we end up [here](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/clusterdeployment_created.go#L82-L84)
> 
> So there seems to have been two options that could have caused continuous creations of integrations: a) `ParseClusterConfig` was constantly erroring b) the `pdData.ServiceID` was never updated: this would be caused by erroring somewhere [during the update/creation of the configmap](https://github.com/openshift/pagerduty-operator/blob/ad70ff6035b4b3f71ec6a959afb996a41f92283c/controllers/pagerdutyintegration/clusterdeployment_created.go#L97-L112)

The issues I witnessed were caused by the `uhc` namespace in hive being deleted:
- The `cm` storing details about the PagerDuty service was already deleted
- But the `cd` was not yet deleting.

The `cm` is used by the operator to know if there is work to be done or not.
- If the `cm` already exist then the operator already created all the the PagerDuty objects and there is nothing to do.
- If the `cm` does not exist (but the `cd` does) then the operator tries to create the PagerDuty objects and stores details about them (mostly their IDs) in the `cm`.

In this scenario the operator tries to reconcile and re-create the PageDuty objects and the `cm` in a loop but never succeeds in writing the `cm` as the namespace is deleting.

While trying to re-create the PagerDuty service is not an issue if the service already exists; re-creating the PagerDuty integration is an issue. Indeed, while it is not possible to create an other PagerDuty service with the same name, creating integration with the same name / type is possible.

You can find more details in [this ticket](https://issues.redhat.com/browse/OHSS-30595).

mjlshen (approved):
/lgtm

PR #271, opened 2024-07-15
make boilerplate-update

mjlshen:
[OSD-24541](https://issues.redhat.com//browse/OSD-24541)


typeid (commented):
@mjlshen should we bump go to 1.22 in the same atomic PR? 

I don't understand why this even builds in the first place with the base image being go1.22 but go.mod being 1.21. Any idea why this works - I believe I've also seen it not work? 🤔

mjlshen:
> should we bump go to 1.22 in the same atomic PR?

There should be no harm and I don't have any reason to not do it, but to explain more:

> I don't understand why this even builds in the first place with the base image being go1.22 but go.mod being 1.21. Any idea why this works - I believe I've also seen it not work? 🤔

It works, Go is pretty strongly backwards compatible. `go.mod`'s [go directive](https://go.dev/ref/mod#go-mod-file-go) describes the language features that are being used in the codebase and sets a minimum version for other codebases to import this module. 

E.g.:
* If a codebase is using `log/slog` that was released in Go 1.21, that would only work if the `go` directive was >= 1.21.
* If a codebase wants to import PDO and we set our `go` directive to 1.21, that codebase must also set their `go` directive to >= 1.21

Neither of those apply to PDO (we're not using any new Go 1.22 features at this time and we don't care about anyone importing this codebase), but you can see an example of holding back on updating the `go` directive in kind: https://github.com/kubernetes-sigs/kind/blob/f28e1072b682c8a30dd8add01cc921124bab7829/go.mod#L3-L11 to minimize impact on other codebases.

typeid (approved):
/lgtm

PR #283, opened 2025-03-11
[draft] bump golang.org/x/net to 0.35.0

anispate:
As part of [OSD-28728](https://issues.redhat.com/browse/OSD-28728) ticket to fix High Vulnerabilities:

GHSA-w32m-9786-jp63 (CVE-2024-45338)
GO-2024-3333 (CVE-2024-45338)

go get golang.org/x/net@v0.35.0
go mod tidy

PR #284, opened 2025-03-12
Updated golang/x/net and /x/net/inda using go get

Mhodesty:
https://quay.io/repository/app-sre/pagerduty-operator/manifest/sha256:c8728bca36387dbdc493f72ab712386bcbd3d109f05e8838833c8c287a2fc0a9?tab=vulnerabilities 

There is a high vulnerability tied to golang.org/x/net, which is fixed in version 0.33.0. I updated the package using go get to that version. It notified me that we also needed to update golang.org/x/net/idna to v 0.33.0 as well, so I did that too. I then built the image and pushed it to my own quay repo to confirm the changes fixed the vulnerability.


Mhodesty:
https://github.com/openshift/pagerduty-operator/pull/283
Closed in favor of this PR which is basically the same thing

PR #329, opened 2025-11-04
SREP-2442: Add distribution of CAD int key via PDIntegration spec

MateSaary:
Keeping in `do-not-merge` until the it's sure that the implementation is right. 

For reviewer context, see:
* https://github.com/openshift/configure-alertmanager-operator/pull/456
* https://issues.redhat.com/browse/SREP-1883


typeid:
It seems a bit overkill to add the CAD integration key via PDO. I think we could directly create it via MCC. 

PDO is mainly there to create & set up new services.

